PR #19, opened 2024-08-14
Add pre-publish preview links

alecgeatches:
## Description

Add "Secure Preview" option to the sidebar of private posts:

secure-link-demo

### Token generation

General token logic was borrowed from the [VIP Decoupled Bundle's `Token` class](https://github.com/Automattic/vip-decoupled-bundle/blob/trunk/preview/token.php). The implementation in this plugin has been changed in a handful of ways:

- Instead of specifying global values for [expiration time](https://github.com/Automattic/vip-decoupled-bundle/blob/7c26592f4651777f13b87b51fef3319f99a9ce4e/preview/token.php#L26) and [whether tokens expire on use](https://github.com/Automattic/vip-decoupled-bundle/blob/7c26592f4651777f13b87b51fef3319f99a9ce4e/preview/token.php#L136), these parameters [are now options](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/token.php#L26) provided to token generation. There's still an existing check that a token lifetime doesn't exceed a maximum amount, but this can be [modified via the `vw_secure_preview_max_token_lifetime` filter](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/token.php#L113) similar to the decoupled implementation.
- Link lifetimes are configured in [`get_link_expiration_options()`](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/secure-preview.php#L121-L149) and can be modified via filter to add shorter or longer preview times. This can be combined with the `vw_secure_preview_max_token_lifetime` filter above to set custom expiration times.
- The logic for one-time-use tokens was determined by filter in the decoupled plugin, but now is determined by individual token. [`validate_token` logic has changed](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/token.php#L73) to handle this stored parameter.

While reviewing, please take special care to ensure token generation and validation looks correct.

### Preview handling

WordPress doesn't provide a straightforward way to allow previewing a non-public post. It could be possible for us to create a new endpoint for secure previews (e.g. `/vip-workflow-preview/<post-id>/<token>`), but then we'd need to ensure the theme, header, and whole page is rendered in the same way as a preview to a logged-in users. In order to avoid rendering a post page from scratch, I used a solution that keeps almost all of the same WordPress flow used for normal previews. The only difference is a `vw-token` parameter used in the URL.

Previews are made possible [via `allow_secure_preview_requests()`](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/secure-preview.php#L86), which hooks `posts_results`.

The [`post_results` filter](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/class-wp-query.php#L3401) is a part of `WP_Query`'s `get_posts()`, fired after a post preview has been queried and loaded in memory, but before user permissions have been checked. The main logic for checking user permissions [are in these lines of `get_posts()`](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/class-wp-query.php#L3455-L3478), with these being the most important ones:

```php
if ( $post_status_obj && ! $post_status_obj->public ) {
	if ( ! is_user_logged_in() ) {
		// User must be logged in to view unpublished posts.
		$this->posts = array();
	} else { /* ... */ }
```

In short, if a post status is non-public and a user isn't logged in, all queried posts are dropped.

To get around this while using a token, we [check that we're in a preview](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/secure-preview.php#L95) then [verify the `vw-token` parameter](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/secure-preview.php#L101) passed to the page is a valid token. If that passes, we mark the post status as `"publish"` for the current query. After the filter completes, the [WordPress check for non-public statuses](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/class-wp-query.php#L3455) is then skipped, and the preview is allowed to load.

### "Copy Link" button

secure-preview-modal

The "Copy Link" button [used in the secure preview modal](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/lib/secure-preview.js#L153-L159) is implemented a little strangely. I've added a lot of comments to the [`CopyFromAsyncButton` class](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/lib/components/copy-from-async-button.js) to explain it, but I'll also add some explanation here.

Gutenberg's main way of exposing the clipboard to plugin developers is the [`useCopyToClipboard()` hook](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-compose/#usecopytoclipboard). The first parameter it takes is a static string or function that can generate a string (`() => string`). This hook generates a ref that can be used with a clickable element, and will copy the given text to the clipboard when that element is clicked.

In our case, when the user clicks "Copy Link", we want to [run a fetch in the background](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/lib/secure-preview.js#L98-L102) based on the link parameters the user selected, get the returned URL, and then copy it to the clipboard. However, this doesn't work with `useCopyToClipboard()`, as promises/async functions [are not handled](https://github.com/WordPress/gutenberg/blob/5dae25b15e82a59d28c6c065b5a3745f3f83980f/packages/compose/src/hooks/use-copy-to-clipboard/index.js#L46-L48) as an option, so we have no way to use an async function and the clipboard together.

The result of trying out some different options lead to the creation of [`CopyFromAsyncButton`](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/lib/components/copy-from-async-button.js), which essentially wraps two buttons. When a user clicks the "Copy Link" button:

1. The main button runs `handleCopyClick()`, which `await`s the async function to run.
2. On completion, the result of the async function is set to the `textToCopy` state.
3. When the component rerenders, [`useLayoutEffect()` is used](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/lib/components/copy-from-async-button.js#L29C2-L34) to click a second, invisible button.
4. This invisible button, [which uses the `useCopyToClipboard() hook`](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/lib/components/copy-from-async-button.js#L24-L27), now has `textToCopy` ready to use synchronously.
5. Finally, the async result is copied to the clipboard and the value is passed back to the parent.

If anyone has a better suggestion on this, please let me know! A good core contribution may be to extend `useCopyToClipboard()` to handle async functions so this wouldn't be necessary.

## Testing

See the GIF at the top of this issue for visual reference on where to find the "Secure Preview" button on the sidebar. There are a lot of scenarios to test, and I'll encourage not following an exact set of steps by not providing them. In general for testing, I have two browsers open, one that's logged-in to create posts, and another not logged-in session to verify preview links. Here are some general features to test:

- Test one-time links only work once.

- Test expiration times work as expected. To do this, you can directly edit [the array in `get_link_expiration_options()`](https://github.com/Automattic/vip-workflow-plugin/blob/7f45ac760a0a30325ca2c4f00351c00881ab99a3/modules/secure-preview/secure-preview.php#L132-L147) or use the `vw_secure_preview_expiration_options` filter to add a shorter timeframe, like a one-minute token. Ensure that the token stops working after one minute.
- Try generating multiple tokens at a time, and ensure they work separately.
- Ensure that the "Secure Preview" button only appears for non-public posts in a custom status. The button should also only appear on post types selected under Admin -> VIP Workflow -> Settings:

    Screenshot 2024-08-15 at 1 21 58 PM

- Ensure preview links can only be generated for posts that have been saved.
- Test that errors during token generation in the REST endpoint display correctly in the editor. You can edit `modules/secure-preview/rest/secure-preview.php` to return a `WP_Error` during generation, and should see that error exposed in the editor.


jsconfig.json
```diff
@@ -4,6 +4,7 @@
 		"module": "ES6",
 		"jsx": "preserve",
 		"checkJs": true,
+		"allowUnreachableCode": true
```
alecgeatches, line 7: Added this because VSCode kept removing code during testing, like if I used an early `return` or commented something out. This doesn't stop ESLint from recognizing the issue, it just stops VSCode from silently obliterating code during a save.

modules/custom-status/lib/custom-status-block.js
```diff
@@ -3,7 +3,7 @@ import './editor.scss';
 import { SelectControl } from '@wordpress/components';
 import { compose } from '@wordpress/compose';
 import { dispatch, select, subscribe, withDispatch, withSelect } from '@wordpress/data';
-import { PluginPostStatusInfo } from '@wordpress/edit-post';
+import { PluginPostStatusInfo } from '@wordpress/editor';
```
alecgeatches: `PluginPostStatusInfo` from `@wordpress/edit-post` was deprecated in WordPress 6.6 and moved to `@wordpress/editor`. This doesn't appear to have any effect on the component.

modules/secure-preview/token.php
```diff
@@ -0,0 +1,122 @@
...
+	 * expiration should be validated. (See validate_token.)
+	 *
+	 * @param  int    $post_id            Post ID.
+	 * @param  string $cap                Capability required to generate this token.
+	 * @param  string $is_one_time_use    Whether this token should be deleted after one use.
+	 * @param  string $expiration_seconds The number of seconds this token should be valid for.
+	 * @return array|WP_Error
+	 */
+	public static function generate_token( $post_id, $cap, $is_one_time_use, $expiration_seconds ) {
+		if ( ! current_user_can( $cap, $post_id ) ) {
```
chriszarate: The required cap should be hardcoded in this function instead of being passed in. There isn't really a use case for anything other than 'edit_posts', and it removes the remote possibility of someone with lower caps gaining access to this function and passing in whatever cap they want. The same change should be made to the decoupled bundle.

modules/secure-preview/token.php
```diff
@@ -0,0 +1,122 @@
...
+		 * lifetime of a token. It provides a way to enforce an upper limit on the
+		 * duration for which a token remains valid.
+		 *
+		 * Understand the security implications of this change:
+		 * the longer the token lifetime, the more security risk you bear with a compromised token.
+		 * Extend the maximum lifetime with caution.
+		 *
+		 * @param int    $default_max_lifetime_seconds The default maximum token lifetime in seconds.
+		 */
+		$max_lifetime_seconds = apply_filters( 'vw_secure_preview_max_token_lifetime', $default_max_lifetime_seconds );
```
chriszarate: I'm not sure I really see the point of a maximum, filterable or not, if all of the possible options are filterable via `vw_secure_preview_expiration_options`. It feels like this entire method is superfluous

modules/secure-preview/token.php
```diff
@@ -0,0 +1,122 @@
...
+	/**
+	 * Validate a token that has been sent with a request.
+	 *
+	 * @param  string $token   Token.
+	 * @param  int    $post_id Post ID.
+	 * @return bool
+	 */
+	public static function validate_token( $token, $post_id ) {
+		$meta_key     = self::META_KEY;
+		$saved_tokens = get_post_meta( $post_id, $meta_key );
```
chriszarate: `false` is the default, but a good signal that we expect multiple
```suggestion
		$saved_tokens = get_post_meta( $post_id, $meta_key, false );
```

modules/secure-preview/secure-preview.php
```diff
@@ -0,0 +1,150 @@
...
+		$is_preview = $query->is_preview() && 1 === count( $posts );
+
+		if ( ! $is_preview ) {
+			return $posts;
+		}
+
+		if ( Token::validate_token( $token, $posts[0]->ID ) ) {
+			// Temporarily set post_status to 'publish' to stop WP_Query->get_posts() from clearing out
+			// unpublished posts before render
+			$posts[0]->post_status = 'publish';
```
chriszarate: Where does the correct status get restored? This feels risky, since other code (and other hooks) may run before it is restored and incorrectly assume the post is published.

modules/secure-preview/rest/secure-preview.php
```diff
@@ -0,0 +1,160 @@
...
+			$post_status = get_post_status( $post_id );
+
+			if ( 'publish' === $post_status ) {
+				return new WP_Error( 'vip-workflow-published-post', __( 'Secure links can not be generated for published posts.', 'vip-workflow' ) );
+			} elseif ( 'auto-draft' === $post_status ) {
+				return new WP_Error( 'vip-workflow-not-custom-status', __( 'Posts must be saved before a secure link can be generated.', 'vip-workflow' ) );
+			} else {
+				return new WP_Error( 'vip-workflow-not-custom-status', __( 'Secure links can only be generated for pre-published posts with a custom status.', 'vip-workflow' ) );
+			}
+		}
```
chriszarate: Given this validation when the token is created, shouldn't the post status be an input to the token envelope and be part of the token validation? In other words, if the post status changes (e.g., the post is published), shouldn't the token be invalidated?

chriszarate (commented):
I left some comments around the token generation and validation. However, I worry about labeling this a "secure preview." 

Visiting a "secure" link generally implies that (WP) user authentication and authorization have taken place, and these bearer tokens are subject to none. Calling them secure may encourage people to select long lifetimes and may contribute to misunderstandings about how they work. 

Can we find a better way to communicate that these bearer tokens allow _anyone_ to view the unpublished post? For example, Google often describes this access model as "Anyone with the link," which neatly describes that the URL itself allows access.

Might we also display a warning with "expire on use" is disabled, especially with a longer token lifetime?

smithjw1:
> Can we find a better way to communicate that these bearer tokens allow _anyone_ to view the unpublished post? For example, Google often describes this access model as "Anyone with the link," which neatly describes that the URL itself allows access.

I think I've been calling this a "secure preview" because that's what customers often call it. I completely agree that it isn't "secure." I think we just drop the word secure and call it a "Preview Link."

> Might we also display a warning with "expire on use" is disabled, especially with a longer token lifetime?

I think some micro copy on the dialog is a good idea:

"Anyone with this link will be able to preview the post."

modules/custom-status/custom-status.php
```diff
@@ -1239,6 +1239,27 @@ public function workflow_is_publish_blocked( $post_id ) {
...
+	 * @return bool True if the post is using a custom status, false otherwise.
+	 */
+	public function is_post_using_custom_status( $post_id ) {
+		$post = get_post( $post_id );
+
+		if ( null === $post ) {
+			return false;
+		}
+
+		$custom_post_types = $this->get_post_types_for_module();
```
ingeniumed: Minor: We should rename this function because it's not module specific anymore.

modules/custom-status/lib/custom-status-block.js
ingeniumed: Would this still be backwards compatible though, since we support all the way back to WordPress 6.0 officially?

modules/secure-preview/lib/secure-preview.js
```diff
@@ -0,0 +1,240 @@
...
+			/>
+		</div>
+	);
+};
+
+const getPostProperties = select => {
+	const { getEditedPostAttribute, getCurrentPostType, getCurrentPost } = select( editorStore );
+
+	const post = getCurrentPost();
+	const isDraft = post?.status === 'auto-draft';
```
ingeniumed: I'd a comment here explaining why this is done this way, and change the name of this var to be `isAutoDraft` instead.

modules/secure-preview/rest/secure-preview.php
```diff
@@ -0,0 +1,160 @@
+<?php
+/**
+ * class EditStatus
```
ingeniumed: Should change this name and the description below

modules/secure-preview/rest/secure-preview.php
ingeniumed: Would echo the same Q

modules/secure-preview/secure-preview.php
```diff
@@ -0,0 +1,150 @@
+<?php
+/**
+ * The secure preview module, for sharing pre-published content.
+ *
+ * @package vip-bundle-decoupled
```
ingeniumed: Would remove this

modules/secure-preview/secure-preview.php
```diff
@@ -0,0 +1,150 @@
+<?php
+/**
+ * The secure preview module, for sharing pre-published content.
```
ingeniumed: Minor spacing issue before the for

modules/secure-preview/token.php
```diff
@@ -0,0 +1,122 @@
...
+		foreach ( $saved_tokens as $saved ) {
+			$is_expired = false;
+
+			// Check for token expiration.
+			if ( ! isset( $saved['expiration'] ) || $current_time > $saved['expiration'] ) {
+				$is_expired = true;
+			}
+
+			// Check if token matches. If it does, mark as expired.
+			if ( ! $is_expired && $token === $saved['token'] && self::TOKEN_ACTION === $saved['action'] ) {
```
ingeniumed: Would be worth expiring the token this way on the status change action that we hook into, in the custom status module.

modules/secure-preview/token.php
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	/**
+	 * Validate a token that has been sent with a request.
+	 *
+	 * @param  string $token   Token.
+	 * @param  int    $post_id Post ID.
+	 * @return bool
+	 */
+	public static function validate_token( $token, $post_id ) {
```
ingeniumed: Would this token show up in the rest api for a post? Guessing no 🤔

modules/secure-preview/secure-preview.php
ingeniumed: Would be curious about this as well.

ingeniumed (changes requested):
- Ran into a weird bug where when the post is first made the timeout value is 1 hour, but if I change the status using the extended status dropdown it changes to 8 hours.
- Should update the README as part of this, to add in a brief explanation of this feature and perhaps a gif as well. This way when we are ready for a beta we don't need to go in and update missing documentation.

modules/custom-status/lib/custom-status-block.js
alecgeatches: @ingeniumed Great point! It looks like this change [was added in WordPress 6.6](https://make.wordpress.org/core/2024/06/18/editor-unified-extensibility-apis-in-6-6/) (from [this PR](https://github.com/WordPress/gutenberg/pull/60814)). I'm thinking we should just bump the WordPress supported version to 6.6, as it'll be what VIP customers are largely using, and we may need to use other 6.1+ features in this plugin for workflow management.

alecgeatches: Did that in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/5116fefd07db581bf8585adf6ac64d953e3ab26e. Let me know if you think that's a bad idea!

modules/secure-preview/rest/secure-preview.php
alecgeatches: > Given this validation when the token is created, shouldn't the post status be an input to the token envelope and be part of the token validation?

Since we have multiple pre-publish statuses, I'm not sure we want to tie a token to a particular custom status. Let's say we're working with a set of custom statuses like this:

1. Pitch
2. Draft
3. Pending review

It would seem valid to me to create a token for "Pitch" and give it to a colleague, and then maybe promote the post to "Draft" in the meantime. Or the same between "Draft" and "Pending review".

I think the more important part is that a token made for a post in a custom status should only be valid while that post has a custom status assigned to it. What do you think about changing `validate_token()` to ensure that the token is only valid for a post using a custom status, and any other statuses are invalid? This gives some flexibility to modify tracking statuses in a workflow without invalidating links, but ensures they only apply to our custom post statuses.

modules/secure-preview/token.php
alecgeatches: That's a good idea for security. Fixed in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/a624afa583a113fdedfd0804ac0c64f1edee9b1c.

modules/secure-preview/token.php
alecgeatches: I agree! Removed in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/a624afa583a113fdedfd0804ac0c64f1edee9b1c.

modules/secure-preview/token.php
alecgeatches: Addressed in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/4a920c7561c0ac05605fd7cf19c22ea9ad72227d. Curiously I also had to add a type check because explicitly adding `false` triggered [a sniff error](https://github.com/Automattic/VIP-Coding-Standards/blob/develop/WordPressVIPMinimum/Sniffs/Functions/CheckReturnValueSniff.php):

Screenshot 2024-08-16 at 11 29 08 AM

modules/custom-status/custom-status.php
alecgeatches: True! Changed to `get_supported_post_types()` in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/f792715e4aa735a2a8119e77fe429f1b5931854e.

modules/secure-preview/lib/secure-preview.js
alecgeatches: Agree that `isDraft` is a confusing name, especially given that we have a "Draft" status. I changed it to `isUnsavedPost` in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/ac739e071f6a077d3c63728c42cf1b639149d23c which makes the usage clearer in the plugin, and also added a comment.

modules/secure-preview/rest/secure-preview.php
alecgeatches: Changed in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/283371c204bacd6e1cf3f6fc2c4d301ed4416f56.

modules/custom-status/lib/custom-status-block.js
ingeniumed: I'd lean towards 6.1 or 6.2 instead, not 6.6 imo

modules/secure-preview/rest/secure-preview.php
ingeniumed: i get what you mean, they are intending for one link to be used throughout the process. In the event that they don't want that, then a one time use token could work or the max time could be shortened.

That said, I'm wondering if this makes sense as a setting that customers could toggle, in case they are conscious of security and want a new token. That way, by default it won't get invalidated but if they want to make it stricter they can. This would be similar to the [cascading/restrictive hierarchy filter we have in the governance plugin](https://github.com/Automattic/vip-governance-plugin?tab=readme-ov-file#vip_governance__is_block_allowed_in_hierarchy).

modules/secure-preview/token.php
alecgeatches: Interesting question! After a bit of research, it appears it will not be available in REST unless we explicitly register the token meta key via [`register_meta()`](https://developer.wordpress.org/reference/functions/register_meta/) and set `'show_in_rest' => true` as an argument. The post type itself must also[ register support for `custom-fields`](https://developer.wordpress.org/reference/functions/add_post_type_support/), so this won't be exposed unless someone jumps through some hoops to do so.

modules/secure-preview/secure-preview.php
alecgeatches: This is a good concern! I looked through the query code and found that after [the `posts_results` filter](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/class-wp-query.php#L3401) is called and [the logic that allows previewing](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/class-wp-query.php#L3452-L3484) is done executing, [`the_posts` filter is called](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/class-wp-query.php#L3558). I added some code to revert the post status using `the_posts`:

https://github.com/Automattic/vip-workflow-plugin/blob/8cf9f782e01f6a174614e306b3a4a2a2994bbe02/modules/preview/preview.php#L101-L124

Most of the other logic in `WP_Query::get_posts()` between `post_results` and `the_posts` is skipped during a preview, so there's not much opportunity to get unexpected side-effects from other parts of WordPress or plugins.

modules/secure-preview/rest/secure-preview.php
alecgeatches: I added some code to explicitly ignore and clear tokens when a preview for a non-custom-status or non-supported post type is used:

https://github.com/Automattic/vip-workflow-plugin/blob/84e5bc42e46db0926a32652a206a47f8d5d0a37f/modules/preview/token.php#L57-L64

Here's how that will work with some different situations:

1. If a preview URL is generated for a post with custom status "Pitch", and then it moves to another custom status "Draft", the key will continue to work.
2. If a post has a token URL generated, and then moves to a non-custom status (either "Publish" or another custom status that isn't managed by our plugin), the key will be rejected and all tokens will be removed.
3. If a post has a token URL generated in a custom status, then briefly moves to "Publish", and then moves back to a custom statues (e.g. "Pending Review"), the URL will continue to work, as long as it wasn't used while the post was in the "Publish" state.

The third situation above may seem a little odd, but I think it's not harmful if this situation occurs. I opted to do key removal in the `validate_token()` function rather than relying on an external `transition_post_status` or similar filter to remove tokens automatically in the background. The main reasoning is that this simplifies the logic - we only need to think about token management in one function, and there are no other side-effects that will also manage tokens. Also, just in case a post doesn't fire the expected hooks (like if a post was updated to `publish` directly in the database), we'll still need this logic in `validate_token()` anyhow.

This solution keeps "security"-related logic in a single flow in one function, which I think is worthwhile.

alecgeatches:
@chriszarate @ingeniumed @smithjw1 Thank you all for your excellent feedback. This is ready for another review! I've responded to individual comments above, but here were the main changes:

- Mentions of a "Secure Link" have been replaced by "Preview Link". I've also added the suggested copy to the link dialog:

    Screenshot 2024-08-19 at 12 47 02 PM

- We now restore a preview post's status after temporarily setting it to `publish`. [See the discussion here](https://github.com/Automattic/vip-workflow-plugin/pull/19#discussion_r1722190181).
- We now explicitly delete tokens when previewing a post outside of the plugin's supported custom statuses. [See the discussion here](https://github.com/Automattic/vip-workflow-plugin/pull/19#discussion_r1722242053).
- Added a "Preview Links" and `vw_preview_expiration_options` filter documentation [to the README](https://github.com/Automattic/vip-workflow-plugin/blob/add/secure-preview-links/README.md).

modules/custom-status/lib/custom-status-block.js
alecgeatches: After internal discussion, I opted to change the minimum down to `6.2` and revert this package name change in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/2d897915463846a3bc83c32386b358c06218009d. Thanks!

modules/preview/lib/preview.js
```diff
@@ -0,0 +1,250 @@
...
+						<div className="vip-workflow-preview-label">{ __( 'Preview', 'vip-workflow' ) }</div>
+
+						{ ! previewUrl && (
+							<Button
+								className="vip-workflow-preview-button"
+								variant="tertiary"
+								size="compact"
+								onClick={ openModal }
+							>
+								{ __( 'Generate Link', 'vip-workflow' ) }
```
ingeniumed, line 63: One thing I noticed: When a one time use token/link is not useable anymore it doesn't vanish from the post, if you are still in the post after generating it. The generate link still shows the old invalid URL. You need to refresh the page to get that to be updated. This is something that isn't specific to our plugin tbh, it's all over Gutenberg.

But I'm flagging it, in case it is easily fixable. It's fine to mark it as a limitation if not

modules/preview/lib/preview.js
```diff
@@ -0,0 +1,250 @@
...
+import { copySmall } from '@wordpress/icons';
+import { store as noticesStore } from '@wordpress/notices';
+import { registerPlugin } from '@wordpress/plugins';
+
+import CopyFromAsyncButton from './components/copy-from-async-button';
+
+/**
+ * Custom component to generate and copy a preview link in the post sidebar.
+ */
+const VIPWorkflowPreview = ( { status, postType, isUnsavedPost } ) => {
```
ingeniumed, line 32: Could you add some more detailed comments in this plugin? It'll make it easier to revisit this, down the line?

ingeniumed (approved):
Minor comments, otherwise it LGTM. Nice work on this 👍🏾

modules/preview/lib/preview.js
alecgeatches: I did an attempt at this in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/77c252587063e3ca1f48d1ddcd498db94ae080a1, but didn't have a lot of comments to add. Most of this code is pretty idiomatic for functional React components, and I'm not sure where most comments will provide context other than re-stating the code. I've already attempted to isolate non-idiomatic code like [`<CopyFromAsyncButton>`](https://github.com/Automattic/vip-workflow-plugin/blob/77c252587063e3ca1f48d1ddcd498db94ae080a1/modules/preview/lib/components/copy-from-async-button.js) in a separate well-documented component.

If you have any specific areas that you feel are lacking, please let me know!

alecgeatches: Added a couple more comments in https://github.com/Automattic/vip-workflow-plugin/pull/19/commits/ff5481791181c42aa4f7a85df8bea2130e5cd746.

ingeniumed (approved):

PR #20, opened 2024-08-21
Add Editorial Metadata support - part 1

ingeniumed:
###  Editorial Metadata

This is the start of the Editorial Metadata module. It's only focused on adding this to wp-admin so one can configure what Editorial Metadata their site will have. Another PR will be coming up, that will be focused on actually making those fields useable in the block editor. In order to avoid making a large PR, and to allow better parallel progress this has been done.

The initial starting metadata fields are the same as Edit Flow, though I've added a todo asking if we should do that or come up with our own. I have shrunk the supported field list down a lot. I figured once the implementation within the block editor was done, and/or we begin opening this up for beta access we can add more in. Again, this can be modified as well as I have added a todo mentioning this.

Beyond that, I have done the rest api differently than the custom status module. The CUD endpoints give back only the modified items, and react handles the updates to the state. Given we don't know how many fields someone might create this felt like the most efficient way to do it, and it's the rest way of doing things. I'd like to make this change to the custom status rest apis as well, and perhaps think about exposing the getAll/getBy apis as well.

As for the UI, I haven't been able to make it 100% responsive for small tablets and phones but for everyone else it works well. So just a heads up when it comes to that.

#### Screenshots

The landing page without hovering

Screenshot 2024-08-21 at 3 33 22 PM

The landing page after hovering, which reveals the buttons

Screenshot 2024-08-21 at 4 06 07 PM

The CU modal

Screenshot 2024-08-21 at 4 06 12 PM

The Delete modal

Screenshot 2024-08-21 at 4 06 20 PM

### Shared Components

I have made a shared components folder and moved all the PHP and JS code that we share to that. This includes the success notice and delete confirmation modal used by custom status. Ideally, the create, and edit modal might get moved there too once we switch the custom status page to use modals instead of cards for it's CUD operations.

### Tests

I ended up adding a PHP Unit testing framework as well. So now we can start writing tests for our code as we work through each feature.

### Misc

I removed a lot of the unused args and methods I could find so we could reduce the remnants of Edit Flow.

## Steps to Test

- Load up the wp-admin page
- Ensure the editorial metadata menu page shows up and it's populated with the same cards as in the screenshot
- Do some CUD operations and ensure everything works as expected
- Also play around with the other parts of the plugin to ensure nothing broke with the files I moved around and renamed


modules/custom-status/custom-status.php
```diff
@@ -540,7 +540,12 @@ public function update_custom_status( $status_id, $args = [] ) {
 	 */
 	public function delete_custom_status( $status_id, $args = [] ) {
 		// Get slug for the old status
-		$old_status_slug = $this->get_custom_status_by( 'id', $status_id )->slug;
+		$old_status = $this->get_custom_status_by( 'id', $status_id );
+		if ( ! $old_status || is_wp_error( $old_status ) ) {
```
alecgeatches: Minor: Looking at `get_custom_status_by( ... )`, it has `WP_Error` as a return in PHPDoc but doesn't appear to actually have any cases to return an error, just `false` on failure. This could be simplified slightly to remove the `is_wp_error()` check and that PHPDoc updated.

modules/custom-status/lib/components/custom-status-editor.js
```diff
@@ -13,7 +13,7 @@ import { useState } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
 import { useEffect } from 'react';
 
-import ConfirmStatusDeleteModal from './modals/confirm-status-delete-modal';
+import ConfirmDeleteModal from '../../../shared/js/components/modals/confirm-delete-modal';
```
alecgeatches, line 16: It'd be cool if we could get one of those `@shared`-type aliases set up for shared components in JS, e.g.

```suggestion
import ConfirmDeleteModal from '@shared/components/modals/confirm-delete-modal';
```

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
```diff
@@ -0,0 +1,181 @@
...
+						} }
+					>
+						{ __( 'Add New Metadata', 'vip-workflow' ) }
+					</Button>
+				</FlexItem>
+				<Flex className="emetadata-items" direction={ [ 'column', 'row' ] } justify={ 'start' }>
+					{ eMetadataTerms.map( eMetadataTerm => {
+						return (
+							<FlexItem className="emetadata-item" key={ eMetadataTerm.term_id }>
+								<Card>
```
alecgeatches, line 127: Do you think you can move this `<Card>` into a new component? Or maybe another way we can to reduce the complexity of this component and split part of it somewhere else.

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
```diff
@@ -0,0 +1,181 @@
...
+		setError( error );
+		setIsModalVisible( false );
+		setIsConfirmingDelete( false );
+		setEMetadataTerm( null );
+	};
+
+	const handleSuccess = ( message, eMetadataTermResult ) => {
+		setError( null );
+		setSuccess( message );
+		if ( eMetadataTerm && ! isConfirmingDelete ) {
```
alecgeatches, line 43: Minor: These lines seem a little close together to me. Could you add some whitespace here and after the final `else` close bracket?

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
```diff
@@ -0,0 +1,181 @@
...
+									</CardHeader>
+								</Card>
+							</FlexItem>
+						);
+					} ) }
+				</Flex>
+			</Flex>
+
+			{ isConfirmingDelete && deleteModal }
+			{ isModalVisible && createEditModal }
```
alecgeatches: `isModalVisible` is unclear as there are a couple of modals that this component uses. How about:

```suggestion
			{ isCreateEditModalVisible && createEditModal }
```

or 

```suggestion
			{ isEditingEditorialMetadata && createEditModal }
```

modules/editorial-metadata/lib/components/modals/create-edit-editorial-metadata-modal.js
```diff
@@ -0,0 +1,98 @@
...
+			size="medium"
+			onRequestClose={ onCancel }
+			closeButtonLabel={ __( 'Cancel', 'vip-workflow' ) }
+		>
+			<TextControl
+				help={ __( 'The name is used to identify the editorial metadata field.', 'vip-workflow' ) }
+				label={ __( 'Name', 'vip-workflow' ) }
+				onChange={ value => {
+					setName( value );
+				} }
```
alecgeatches: Shortcut:

```suggestion
				onChange={ setName }
```

Same a couple of places below as well.

modules/editorial-metadata/lib/editorial-metadata-configure.js
```diff
@@ -0,0 +1,24 @@
...
+			<EditorialMetadataManager
+				supportedMetadataTypes={ VW_EDITORIAL_METADATA_CONFIGURE.supported_metadata_types }
+				editorialMetadataTerms={ VW_EDITORIAL_METADATA_CONFIGURE.editorial_metadata_terms }
+			/>
+		);
+	}
+} );
+
+if ( module.hot ) {
+	module.hot.accept();
```
alecgeatches, line 23: @ingeniumed Have you been able to use hot reloading successfully?

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
```diff
@@ -0,0 +1,181 @@
...
+									<CardHeader>
+										<Flex direction={ [ 'column' ] } justify={ 'start' } align={ 'start' }>
+											<Heading level={ 4 }>{ eMetadataTerm.name }</Heading>
+											<Text>
+												<i>{ eMetadataTerm.description }</i>
+											</Text>
+										</Flex>
+										<Flex direction={ [ 'column', 'row' ] } justify={ 'end' } align={ 'end' }>
+											<div className="crud-emetadata">
+												<Button
```
alecgeatches, line 139: May want to try `size="compact"` for these buttons

alecgeatches (changes requested):
Thank you for adding tests! And migrating a lot of the editorial metadata system so quickly.

I left a few comments above. The main code change I'd like is to simplify `<EditorialMetadataManager>` if possible. There's a lot of state and deeply-nested markup that could probably be made easier to read with one or two sub-components.

Additionally, I noticed some UI quirks testing locally:

1. The cards appear to have different heights at times. See the "Assignment" card below. This occurred when I was using my 1080p screen with developer tools open:

    smooshed-cards

2. I also noticed some card issues while testing in mobile mode, as you mentioned in the description:

    little-mobile-cards

    This screenshot was taken particularly with the "iPhone 12/13 Pro Max iOS 14.6" profile, but I saw similar results with other mobile sizes.

    Additionally, we should probably have the hover states removed when on mobile and show the buttons all the time.

3. More of a question. It feels a little weird to me that the statuses aren't reorderable. We could certainly add that behavior later, but I found myself wanting to change the order in the UI. Maybe that would lend itself better to a fully column-based layout?

modules/custom-status/lib/components/custom-status-editor.js
ingeniumed: For some reason, I thought that was ts or next.js specific lemme dig into that a bit. That would be a great QOL improvement.

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
ingeniumed: I get what you mean, as there is a lot going on here. Lemme try to attempt that and see what I can do

modules/custom-status/custom-status.php
ingeniumed: Huh, you are right. I missed that while trying to re-work those methods. It's weird that it's giving back a false on failure or an actual value on true. Lemme see where all it's used to make sure it's correctly checked and I'll update the docs too

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
ingeniumed: Good point, I'll make that change.

modules/editorial-metadata/lib/editorial-metadata-configure.js
ingeniumed: Yip, I was able to!

ingeniumed:
> The cards appear to have...and show the buttons all the time.

Agreed with all of this, as I noticed it as well. Lemme see what more I can do to make this better. I had tried to perfect this as much as I could, and then put this PR up so I could get feedback and solutions on how best to improve this.

> More of a question. It feels a little weird to me that the statuses aren't reorderable. We could certainly add that behavior later, but I found myself wanting to change the order in the UI. Maybe that would lend itself better to a fully column-based layout?

That's an interesting observation as my goal had been to make it not reorderable, which is why I made it this way. Since the order in this case doesn't matter, their value is really what matters. I was trying to make the layout be closer to a table, by being column based and then be row based depending on what fits on every row.

modules/custom-status/lib/components/custom-status-editor.js
ingeniumed: Scratch that, I realized that I did try and get this to work using `jsconfig` and `webpack.config.js` but kept running into build problems so I abandoned that for now. Can come back to this later though I 100% agree that this is very useful to have

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
ingeniumed: This made me realize that the error notice during an edit, and add operation should be raised within the modal dialog itself. It shouldn't go on the main page, as then you need to re-enter all your info. Besides that, I tried to simplify this down a touch by making a common error notice. 

I tried to move the `<Card>` to a new component, but in order to get it working the complexity is nearly the same. This is because the API calls need to be made, and the resulting state still needs to be managed. When we switch the custom status to be modal dialog based, I suspect there will be some common components made and that point we could reduce this down a touch. I'll leave this as is for now.

ingeniumed:
I re-did the UI so that all those UI glitches don't happen again. What I changed this time is that I stopped hiding the button and only showing them on hover. Instead, I made them into a pencil and trash and made them always visible. By using those icons, they aren't as distracting and fit right in. We can change the colour scheme to grayscale if need be, but that would make it inconsistent with the rest of the plugin imo. 

Check it out and lemme know what you think.

This is in 1080p.

Screenshot 2024-08-22 at 4 11 55 PM

This is for an iPhone 11.

Screenshot 2024-08-22 at 4 12 05 PM

This is for my native 4K monitor.

Screenshot 2024-08-22 at 4 12 18 PM

ingeniumed:
Pushed up all the necessary changes, it should be good to review once more.

Note that the icons are provided by Gutenberg itself - https://wordpress.github.io/gutenberg/?path=/docs/icons-icon--docs.

alecgeatches (approved):
Looking good to me!

PR #21, opened 2024-08-21
Add Tracks integration

hanifn:
## Description

Add Tracks integration into the plugin. This is dependent on the following MU-plugins PR: https://github.com/Automattic/vip-go-mu-plugins/pull/5770

## Steps to Test

1. Check out the MU-plugin PR above.
2. Check out this PR into the `plugins` directory of a local [VIP Go Skeleton repo](https://github.com/Automattic/vip-go-skeleton)
3. Create a new VIP dev-env with `vip dev-env --slug workflow-test --app-code={path to vip-go-skeleton repo} --mu-plugins={path to MU-plugins repo}`
4. Start the dev-env with `vip dev-env --slug workflow-test start`
5. Test the plugin


class-workflow.php
```diff
@@ -120,6 +121,8 @@ private function load_modules() {
 	private function setup_actions() {
 		add_action( 'init', [ $this, 'action_init' ] );
 		add_action( 'init', [ $this, 'action_init_after' ], 1000 );
+		// Load the telemetry module in init as all plugins are loaded by then
+		add_action( 'init', [ $this, 'start_tracking' ] );
```
alecgeatches: Maybe `init_telemetry` as a function name alternative?

modules/telemetry/class-vip-workflow-tracker.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace VIPWorkflow\Modules\Telemetry;
```
alecgeatches: Edit: Incorrect, see below.

~I think this is a little confusing as a "module", given that the other modules extend `Module` and have a different initialization pattern. However, we are looking to change how the module system works soon, and I think this better reflects the low-boilerplate direction we want to go in. So I'm good with this for now!~

alecgeatches: Nevermind, I see the actual `Telemetry` class below, which is a proper module.

class-workflow.php
alecgeatches: Can this function be called directly in the telemetry module's `init()`? It would seem more relevant in there.

modules/telemetry/events/notification-events.php
```diff
@@ -0,0 +1,20 @@
...
+use WP_Post;
+use WP_User;
+
+function record_notification_sent(
+	WP_Post $post,
+	string $subject,
+	WP_User $user,
+	Tracker $tracker
+) {
+	$tracker->record_event( 'notification_sent', [
```
alecgeatches: I think we can remove some PII here, maybe even all of it:

- `$subject` can contain user display names and email addresses, which I don't think is necessary for us to track.
- `$post` is an entire WP_Post object which seems like a lot. I was curious what that looked like, here's an example output from `json_encode( (array) get_post( <post_id> ) )`:

    ```json
    {
      "ID": 30303,
      "post_author": "1",
      "post_date": "2024-08-27 19:35:04",
      "post_date_gmt": "0000-00-00 00:00:00",
      "post_content": "<!-- wp:paragraph -->\n<p>Inner content here</p>\n<!-- /wp:paragraph -->",
      "post_title": "fresh post",
      "post_excerpt": "",
      "post_status": "last-pendingreview",
      "comment_status": "open",
      "ping_status": "open",
      "post_password": "",
      "post_name": "",
      "to_ping": "",
      "pinged": "",
      "post_modified": "2024-08-27 19:35:04",
      "post_modified_gmt": "2024-08-27 19:35:04",
      "post_content_filtered": "",
      "post_p arent": 0,
      "guid": "http:/workflow.vipdev.lndo.site/?p=30303",
      "menu_order": 0,
      "post_type": "post",
      "post_mime_type": "",
      "comment_count": "0",
      "filter": "raw"
    }
    ```

    There's a lot of PII in here including the entire `post_content`, `post_excerpt`, and potentially others. I think the most we should track would be a post ID.

- `$user`. Here's a `json_encode( (array) get_user_by( 'id', 1 ) )` for a default `vip dev-env` user:

    ```json
    {
        "data": {
            "ID": "1",
            "user_login": "vipgo",
            "user_pass": "$P$BShVqNoDOqa\/ukp4s.xwYO85QDsIx81",
            "user_nicename": "vipgo",
            "user_email": "vip@localhost.local",
            "user_url": "http://workflow.vipdev.lndo.site",
            "user_registered": "2024-07-03 15:25:18",
            "user_activation_key": "",
            "user_status": "0",
            "display_name": "vipgo"
        },
        "ID": 1,
        "caps": {
            "administrator": true,
            "view_query_monitor": true
        },
        "cap_key": "wp_capabilities",
        "roles": [
            "administrator"
        ],
        "allcaps": {
            "switch_themes": true,
            "edit_themes": true,
            "activate_plugins": true,
            "edit_plugins": true,
            "edit_users": true,
            "edit_files": true,
            "manage_options": true,
            "moderate_comments": true,
            "manage_categories": true,
            "manage_links": true,
            "upload_files": true,
            "import": true,
            "unfiltered_html": true,
            "edit_posts": true,
            "edit_others_posts": true,
            "edit_published_posts": true,
            "publish_posts": true,
            "edit_pages": true,
            "read": true,
            "level_10": true,
            "level_9": true,
            "level_8": true,
            "level_7": true,
            "level_6": true,
            "level_5": true,
            "level_4": true,
            "level_3": true,
            "level_2": true,
            "level_1": true,
            "level_0": true,
            "edit_others_pages": true,
            "edit_published_pages": true,
            "publish_pages": true,
            "delete_pages": true,
            "delete_others_pages": true,
            "delete_published_pages": true,
            "delete_posts": true,
            "delete_others_posts": true,
            "delete_published_posts": true,
            "delete_private_posts": true,
            "edit_private_posts": true,
            "read_private_posts": true,
            "delete_private_pages": true,
            "edit_private_pages": true,
            "read_private_pages": true,
            "delete_users": true,
            "create_users": true,
            "unfiltered_upload": true,
            "edit_dashboard": true,
            "update_plugins": true,
            "delete_plugins": true,
            "install_plugins": true,
            "update_themes": true,
            "install_themes": true,
            "update_core": true,
            "list_users": true,
            "remove_users": true,
            "promote_users": true,
            "edit_theme_options": true,
            "delete_themes": true,
            "export": true,
            "administrator": true,
            "view_query_monitor": true
        },
        "filter": null,
        "\u0000WP_User\u0000site_id": 1
    }
    ```

    Clearly that's a lot of excess data, and includes a lot of PII like emails and security-related data like password hashes.

I think we should just send:

```php
[
    'post' => $post->ID
]
```

and that'll be enough for us to understand how much the notification feature is being used.

modules/telemetry/events/settings-events.php
```diff
@@ -0,0 +1,64 @@
...
+
+use VIPWorkflow\Modules\Telemetry\Tracker;
+use VIPWorkflow\VIP_Workflow;
+
+function record_admin_update( string $previous_version, string $new_version, Tracker $tracker ): void {
+	// Get all custom statuses
+	$custom_statuses = VIP_Workflow::instance()->custom_status->get_custom_statuses();
+
+	// Get all posts count
+	$posts_count = wp_count_posts();
```
alecgeatches: Cool function! I tried this locally with a DB containing ~30k posts, and this was approximately instant.

alecgeatches: Note that [`wp_count_posts()`](https://developer.wordpress.org/reference/functions/wp_count_posts/) only counts `post`-types by default. We should probably loop through each supported post type and sum counts for each status.

modules/telemetry/events/settings-events.php
```diff
@@ -0,0 +1,64 @@
...
+	// Only care about published and posts with custom status
+	$total_posts = (int) $posts_count->publish;
```
alecgeatches: I'm curious why published posts' count is summed with custom statuses. It seems like a customer with 50 custom status posts and 50 published posts would have the same event data as a customer with 0 custom status posts and 100 published posts.

Maybe we can send `total_posts` and `custom_status_posts` counts as separate pieces of event data?

modules/telemetry/events/settings-events.php
```diff
@@ -0,0 +1,64 @@
...
+	] );
+}
+
+function record_settings_update( array $new_options, array $old_options, Tracker $tracker ): void {
+	if ( $new_options['publish_guard'] !== $old_options['publish_guard'] ) {
+		record_publish_guard_toggle( $new_options['publish_guard'], $tracker );
+	}
+
+	if ( $new_options['send_to_webhook'] !== $old_options['send_to_webhook'] ) {
+		record_send_to_webhook_toggle( $new_options['send_to_webhook'], $new_options['webhook_url'], $tracker );
```
alecgeatches: `webhook_url` may contain private information like an auth key, right? We should probably remove this from the event data.

modules/telemetry/events/status-events.php
```diff
@@ -0,0 +1,62 @@
...
+use VIPWorkflow\Modules\Telemetry\Tracker;
+use WP_Post;
+
+function record_custom_status_change(
+	string $new_status,
+	string $old_status,
+	WP_Post $post,
+	Tracker $tracker
+): void {
+	if ( 'post' !== $post->post_type ) {
```
alecgeatches: We should also check `get_supported_post_types()` here.

modules/telemetry/events/status-events.php
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function record_add_custom_status(
+	string $term,
+	array $args,
+	Tracker $tracker
+): void {
+	$tracker->record_event( 'custom_status_created', [
+		'term' => $term,
+		'args' => $args,
```
alecgeatches: If I'm following the code correctly `$args` can contain a term description which is extra data we probably don't want. Can we just send the term ID?

modules/telemetry/events/status-events.php
```diff
@@ -0,0 +1,62 @@
...
+function record_delete_custom_status(
+	int $status_id,
+	string $slug,
+	array $args,
+	Tracker $tracker
+): void {
+	$tracker->record_event( 'custom_status_deleted', [
+		'status_id' => $status_id,
+		'slug'      => $slug,
+		'args'      => $args,
```
alecgeatches: Same thought here as above: https://github.com/Automattic/vip-workflow-plugin/pull/21/files#r1733537328.

modules/telemetry/events/status-events.php
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function record_update_custom_status(
+	int $status_id,
+	array $args,
+	Tracker $tracker
+): void {
+	$tracker->record_event( 'custom_status_changed', [
+		'status_id' => $status_id,
+		'args'      => $args,
```
alecgeatches: Also probably want to remove `$args` here.

alecgeatches (changes requested):
I left a few comments above, mostly about removing PII in telemetry data (like `WP_Post` or `WP_User` objects).

I do have an overall question about the indirection we're using to send events. For example, let's look at the event when the plugin's version updates. First, we fire an action during update:

https://github.com/Automattic/vip-workflow-plugin/blob/8c7f9ded80565a9e764bc7041c0db0357c002fbb/class-workflow.php#L167

In the telemetry module, we catch that action and send it to `Tracker::track_event()`:

https://github.com/Automattic/vip-workflow-plugin/blob/8c7f9ded80565a9e764bc7041c0db0357c002fbb/modules/telemetry/telemetry.php#L62-L67

`track_event` then calls the passed callback, getting passed arguments though `func_get_args()`, and calling `VIPWorkflow\Modules\Telemetry\Events\record_admin_update` indirectly:

https://github.com/Automattic/vip-workflow-plugin/blob/8c7f9ded80565a9e764bc7041c0db0357c002fbb/modules/telemetry/class-vip-workflow-tracker.php#L41-L53

Lastly, `record_admin_update()` is called:

https://github.com/Automattic/vip-workflow-plugin/blob/8c7f9ded80565a9e764bc7041c0db0357c002fbb/modules/telemetry/events/settings-events.php#L8

which uses the injected `$tracks` object to send an event:

https://github.com/Automattic/vip-workflow-plugin/blob/8c7f9ded80565a9e764bc7041c0db0357c002fbb/modules/telemetry/events/settings-events.php#L20-L25

---

This is a lot of indirection, and a lot of places that need to be changed to add a new data point. What if we had the `Telemetry` module register the action and also fire it off? e.g.

First, we fire the `do_action` from settings code as normal:

https://github.com/Automattic/vip-workflow-plugin/blob/8c7f9ded80565a9e764bc7041c0db0357c002fbb/class-workflow.php#L167

Then in `telemetry/telemetry.php`, we register a handler:

```php
add_action( 'vw_upgrade_version', [ $this, 'track_upgrade_version' ], 10, 2 )
```

and in `track_upgrade_version`, we send the event:

```php
public function track_upgrade_version( $previous_version, $new_version ) {
    // ...
    $this->tracker->record_event( 'administration_update', [ 
        'previous_version' => $previous_version, 
        'new_version'      => $new_version, 
        'custom_statuses'  => count( $custom_statuses ), 
        'total_posts'      => $total_posts, 
    ] ); 
}
```

Removing some of the abstractions just seems easier to follow for me. What do you think?

modules/telemetry/class-vip-workflow-tracker.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	/**
+	 * Record an event
+	 *
+	 * @param string $event_name The event name
+	 * @param array $event_data The event data
+	 */
+	public function record_event( string $event_name, array $event_data = [] ): void {
+		$this->tracks->record_event( $event_name, $event_data );
```
alecgeatches: Does this all work properly outside of a VIP site? I see there's a class check in [`start_tracking`](https://github.com/Automattic/vip-workflow-plugin/blob/8c7f9ded80565a9e764bc7041c0db0357c002fbb/class-workflow.php#L389), but how do event calls work on a non-VIP site?

class-workflow.php
hanifn: I wanted to separate the checks for the `Tracks` class and the actual telemetry implementation. We also need to make sure the MU plugins telemetry lib is loaded _before_ loading the telemetry module here.

modules/telemetry/events/notification-events.php
hanifn: Yes that makes sense. Outside of PII concerns, that's also just _a lot_ of data that isn't really contextually needed.

hanifn: Removed in 1484e37eb986f780babb89c527958022fbe395e2

modules/telemetry/events/status-events.php
hanifn: Done in 1484e37eb986f780babb89c527958022fbe395e2

modules/telemetry/events/status-events.php
hanifn: Done in 1484e37eb986f780babb89c527958022fbe395e2

modules/telemetry/events/status-events.php
hanifn: Done in 1484e37eb986f780babb89c527958022fbe395e2

modules/telemetry/events/settings-events.php
hanifn: Removed in 82596947dc3c7aa86b53b8b109219e48862fad47

modules/telemetry/events/status-events.php
hanifn: Done in 5752761a3e7e03a51257369550656b7e257602db

modules/telemetry/class-vip-workflow-tracker.php
hanifn: Unless they've installed the VIP MU-plugins on their non-VIP site then the `Telemetry::init()` won't be called in `start_tracking` due to the class check you've mentioned. If they did install VIP MU-plugins on their non-VIP site, then events should still be recorded but the user ID and user ID type would be different as those depends on VIP env constants

modules/telemetry/events/settings-events.php
hanifn: Done in a2dc1b5fefe29f51c89f84a7a2093ea4f466d251

hanifn:
> Removing some of the abstractions just seems easier to follow for me. What do you think?

@alecgeatches I agree. Your explanation really cleared up how many levels of abstractions there were in my original attempt. I've refactored it in 4db9b2b92941572b482ad7a9aabfe86fa811e93b and that should be a little more straightforward now

modules/telemetry/events/settings-events.php
```diff
@@ -0,0 +1,108 @@
...
+		// Get supported post types
+		$supported_post_types = VIP_Workflow::instance()->custom_status->get_supported_post_types();
+
+		$total_posts = 0;
+		foreach ( $supported_post_types as $post_type ) {
+			// Get all posts count for each post type
+			$posts_count = wp_count_posts( $post_type );
+
+			// Only care about published and posts with custom status
+			$total_posts += (int) $posts_count->publish;
```
alecgeatches: @hanifn Still curious about this data point [RE this reply](https://github.com/Automattic/vip-workflow-plugin/pull/21#discussion_r1733529489) from last week.

My basic problem is that if we combine `publish` and custom status post counts together, the posts in custom statuses will likely be overwhelmed by published post counts, and then we'll essentially lose data about custom statuses. Why include the count of published posts at all? And if we want that data for a reason, how about also including `custom_status_posts` as a separate count?

modules/telemetry/events/status-events.php
```diff
@@ -0,0 +1,104 @@
...
+	 * Record an event when a custom status is created
+	 *
+	 * @param string $term The term name
+	 * @param string $slug The term slug
+	 * @param array $args The term arguments
+	 */
+	public function record_add_custom_status(
+		string $term,
+		string $slug,
+		array $args,
```
alecgeatches: For each of these, can we remove the `$args` parameter, especially if it's not being used?

modules/telemetry/events/status-events.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @param int $status_id The status ID
+	 * @param array $args The status arguments
+	 */
+	public function record_update_custom_status(
+		int $status_id,
+		array $args,
+	): void {
+		$this->tracks->record_event( 'custom_status_changed', [
+			'status_id' => $status_id,
+			'slug'      => $args['slug'],
```
alecgeatches: For this `$args` term, can we pass `$slug` directly in the `vw_update_custom_status` action call?

https://github.com/Automattic/vip-workflow-plugin/blob/3a369543e0e36cb0a155329ded3fb8eb560e2197/modules/custom-status/custom-status.php#L551

It seems like the action-calling code will better know what information to extract from `$args` instead of the inner telemetry code here. We can always update the action call if we want to change those parameters.

modules/telemetry/telemetry.php
```diff
@@ -0,0 +1,84 @@
...
+		$this->tracks = new Tracks( 'vip_workflow_' );
+	}
+
+	/**
+	 * Initialize the module and register event callbacks
+	 */
+	public function init(): void {
+		// Add custom status events
+		$status_events = new Status_Events( $this->tracks );
+		add_action(
```
alecgeatches: Minor: I like that you've split the code up to gather events into separate classes, but it could be even better separated if we put the action registration calls in those classes as well, as `Status_Events`' functions are tightly bound to the action registration. You don't need to do this if you disagree.

alecgeatches (approved):
@hanifn Thank you for the simplifying changes! I requested a few edits above, but nothing major. We should probably wait until https://github.com/Automattic/vip-go-mu-plugins/pull/5770 is merged in case there are any last-minute function API changes, but otherwise I'm good to merge. Thank you!

modules/telemetry/telemetry.php
hanifn: That's a good idea actually. Changed in a0f9415fec276b3acf490773e9dc9db2548f8d36

modules/telemetry/events/status-events.php
hanifn: The `do_action` call is passing those arguments though so wouldn't the callback be required to accept it event if it's not used?

modules/telemetry/events/settings-events.php
hanifn: > statuses will likely be overwhelmed by published post counts, and then we'll essentially lose data about custom statuses. Why include the count of published posts at all? And if we want that data for a reason, how about also including custom_status_posts as a separate count?

@alecgeatches Yes I see your point. I think adding the `custom_status_posts` would also be useful. Would additional comparison points be more useful? I'm thinking adding `custom_status_posts` and `published_posts` on top of the existing `total_posts`

modules/telemetry/telemetry.php
alecgeatches: Thank you, looks great!

modules/telemetry/events/status-events.php
alecgeatches: I'm not sure I understand. In WordPress you're allowed to accept fewer arguments than the callback supplies, so you can register to receive just the first two arguments if desired.

However, I'm just thinking we change the `do_action()` calls to not supply a full `$args` array unless we're recording all of the `$args`, or to only supply targeted data like [suggested below](https://github.com/Automattic/vip-workflow-plugin/pull/21#discussion_r1742441009). The main problems with supplying unnecessary `$args` is:

- If the `do_action()` calling code is modified or refactored, `$args` might not keep the same structure the telemetry part of the code expects. For example when we call the `vw_add_custom_status` action here, we're passing `$args` that can contain `'slug'`, `'position'`, and any other optional arguments given to the function. If that code is refactored in the future, the shape of `$args` might change, which is also coupled with telemetry code, which may also need to be changed. However, if we just pass a `$slug` parameter directly from the `vw_add_custom_status` action call, there's less coupling and room for error.
- Passing unnecessary data to telemetry functions that could potentially contain PII doesn't look very safe, even if we just internally discard most of that data. We should be careful and selective about the exact data we're passing in.

modules/telemetry/events/settings-events.php
alecgeatches: If `total_posts === custom_status_posts + published_posts`, then we probably just need the two terms on the right side and I say we drop `total_posts`. But in general, if we're already doing the queries we need to count posts it doesn't really hurt to send more counts.

modules/telemetry/events/status-events.php
hanifn: > I'm not sure I understand. In WordPress you're allowed to accept fewer arguments than the callback supplies, so you can register to receive just the first two arguments if desired.

Ah thanks for clearing that up. I forgot that that was possible 👍 

> However, I'm just thinking we change the do_action() calls to not supply a full $args array unless we're recording all of the $arg

My thinking when I added the `do_action` calls was that we should also ensure that these hooks are also useful for things other than telemetry(i.e.: expanding the notifications module) thus I wanted to include as much contextual data as possible. That said, I see your point that the `$args` argument could be inconsistent and it would be better to be more specific. I'll make the changes.

hanifn: Done in 07fb1468b736c69b9b9c6a5ef025ee36b0ccd741

modules/telemetry/events/status-events.php
hanifn: Changed in 07fb1468b736c69b9b9c6a5ef025ee36b0ccd741

modules/telemetry/events/settings-events.php
hanifn: Agreed. Changed in c9ac0647a7a79da1316e03f9d6eb5408ffdc094e

alecgeatches: It looks like we're not actually using `$custom_status_posts`, is that right?

modules/telemetry/events/status-events.php
alecgeatches: Thank you! It still looks like we're passing an `$args` along with `vw_add_custom_status` and `vw_delete_custom_status`, can you get those more specific as well?

hanifn: Done in 6a58f5ae591b571ecb8933c58ff152776a7d0f5b

modules/telemetry/events/settings-events.php
hanifn: > It looks like we're not actually using `$custom_status_posts`, is that right?

Oops that's a mistake. Fixed in 163daded7fb13b2408c548bda8c2543ffb2aa186

modules/notifications/notifications.php
```diff
@@ -147,13 +147,36 @@ public static function notification_status_change( string $new_status, string $o
 
 			$action = 'status-change';
 
-			self::schedule_emails( $action, $post, $subject, $body );
+			$notification_email = OptionsUtilities::get_options_by_key( 'email_address' );
```
alecgeatches, line 150: @ingeniumed The notification code has been modified to check for email/webhook presence in this upper function. We use this data (`$is_email_scheduled` and `$is_webhook_scheduled`) to determine if we should trigger `vw_notification_status_change` telemetry at all, and also to include which notification types were sent in that datapoint.

modules/telemetry/telemetry.php
```diff
@@ -0,0 +1,232 @@
...
+			// The status hasn't changed
+			return;
+		} elseif ( in_array( $new_status, [ 'inherit', 'auto-draft' ] ) ) {
+			// The status hasn't changed, or it moved into an auto-generated status
+			return;
+		}
+
+		$status_slugs = wp_list_pluck( CustomStatus::get_custom_statuses(), 'slug' );
+
+		if ( ! in_array( $new_status, $status_slugs, true ) && ! in_array( $old_status, $status_slugs, true ) ) {
```
alecgeatches, line 62: The logic in this function has been modified so that we'll still track if a post moves from `some-custom-status` to `publish`. If the prior or current status is custom, we track.

modules/telemetry/telemetry.php
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	/**
+	 * Record an event when a custom status is updated
+	 *
+	 * @param WP_Term $updated_status The updated status WP_Term object.
+	 * @param array $update_args The arguments used to update the status.
+	 */
+	public static function record_update_custom_status( WP_Term $updated_status, array $update_args ): void {
+		$is_position_update = 1 === count( $update_args ) && isset( $update_args['position'] );
```
alecgeatches, line 115: Here we use the passed-in `$update_args` to determine if there's just a position switch happening. Before this code was added, any reorder caused an event to be tracked for every status.

modules/telemetry/telemetry.php
```diff
@@ -0,0 +1,232 @@
...
+
+		$custom_status_posts = 0;
+
+		foreach ( $supported_post_types as $post_type ) {
+			// Get all posts count for this post type
+			$posts_count = wp_count_posts( $post_type );
+
+			foreach ( $custom_statuses as $status ) {
+				if ( isset( $posts_count->{ $status->slug } ) ) {
+					$custom_status_posts += (int) $posts_count->{ $status->slug };
```
alecgeatches: This event has been modified to only count posts in a custom status. Previously we were also tracking published post counts, but I didn't see the need for that datapoint.

modules/telemetry/telemetry.php
```diff
@@ -0,0 +1,232 @@
...
+		if ( $new_options['publish_guard'] !== $old_options['publish_guard'] ) {
+			self::record_publish_guard_toggle( $new_options['publish_guard'] );
+		}
+
+		if ( $new_options['webhook_url'] !== $old_options['webhook_url'] ) {
+			self::record_send_to_webhook_toggle( $new_options['webhook_url'] );
+		}
+
+		if ( $new_options['email_address'] !== $old_options['email_address'] ) {
+			self::record_send_to_email_toggle( $new_options['email_address'] );
```
alecgeatches, line 202: Added this datapoint for tracking when a custom email address is added/deleted/changed, which works the same as the webhook URL toggle tracking.

modules/telemetry/telemetry.php
smithjw1: I think we want "published" as well for a few reasons:

* Since each of these events has a timestamp, it might show associations in workflow shapes, complications, and publishing velocity.
* If we have a site with zero published posts, we know something is wonky.
* If we started to record Post IDs we could track the time it takes for a post to go from start to finish.

Maybe this is low value/far out, but I'm interested in others thoughts.

alecgeatches: I think that's a good reason, I'll add it back in.

ingeniumed: Agreed on that. This would also help us better support complicated workflows, and would be beneficial down the line to target new customers who haven't used WordPress before as they are fond of Adobe, etc instead.

modules/custom-status/custom-status.php
```diff
@@ -32,11 +32,11 @@ class CustomStatus {
 	const SETTINGS_SLUG = 'vw-custom-status';
 
 	// The metadata keys for the custom status term
-	const METADATA_POSITION_KEY = 'position';
+	const METADATA_POSITION_KEY          = 'position';
```
ingeniumed, line 35: The spacing is interesting as I would have thought running `composer run cbf` should catch that. Anything that you think would help with this? Prettier maybe on php files as well?

modules/custom-status/custom-status.php
```diff
@@ -534,14 +534,23 @@ public static function add_custom_status( array $args ): WP_Term|WP_Error {
...
+		$term_id = $inserted_term['term_id'];
+
+		/**
+		 * Fires after a custom status is added to the database.
+		 *
+		 * @param int $term_id The ID of the custom status.
+		 * @param string $term_name The name of the custom status.
+		 * @param string $slug The slug of the custom status.
+		 */
+		do_action( 'vw_add_custom_status', $term_id, $term_name, $term_to_save['slug'] );
```
ingeniumed: I'd actually not put this here. This should be on line 571 instead, right before the `get_custom_status_by` call. I've tried to make the whole process transactional:

- Save the status
- Save each of the metadata info as `term_meta` linked to the status
- If any step fails, delete all the bad data

So, keeping this here would mess the event up if the status is deleted as bad data. Plus, we would be able to potentially have the entire status object be sent to the event tracking code. That would be useful for recording what type of EM fields are usually linked, etc.

modules/custom-status/custom-status.php
```diff
@@ -647,9 +656,17 @@ public static function update_custom_status( int $status_id, array $args = [] ):
...
-		$status_result = self::get_custom_status_by( 'id', $status_id );
+		$updated_status = self::get_custom_status_by( 'id', $status_id );
+
+		/**
+		 * Fires after a custom status is updated in the database.
+		 *
+		 * @param WP_Term $updated_status The updated status WP_Term object.
+		 * @param array $update_args The arguments used to update the status.
+		 */
+		do_action( 'vw_update_custom_status', $updated_status, $args );
```
ingeniumed: Same thing as add in that, we should provide the entire status imo. That way we can get access to the EM fields linked as well.

modules/notifications/notifications.php
```diff
@@ -223,8 +240,8 @@ public static function schedule_emails( string $action, WP_Post $post, string $s
 		 */
 		$message_headers = apply_filters( 'vw_notification_email_headers', $message_headers, $action, $post );
 
-		if ( ! empty( $email_recipients ) ) {
-			wp_schedule_single_event( time(), 'vw_send_scheduled_emails', [ $email_recipients, $subject, $message, $message_headers ] );
+		if ( ! empty( $recipients ) ) {
```
ingeniumed: We should switch the use of `empty` for the actual `[]` check. 

Hat tip to Alexey for that.

modules/notifications/notifications.php
```diff
@@ -274,6 +286,11 @@ public static function schedule_webhook_notification( string $webhook_message, s
 	public static function send_to_webhook( string $message, string $message_type, string $timestamp ): bool {
 		$webhook_url = OptionsUtilities::get_options_by_key( 'webhook_url' );
 
+		if ( empty( $webhook_url ) ) {
```
ingeniumed: Same thing here.

I know this is code that's just been moved but figured it's good to target it now.

vip-workflow.php
```diff
@@ -62,8 +70,16 @@
 require_once VIP_WORKFLOW_ROOT . '/modules/shared/php/util.php';
 require_once VIP_WORKFLOW_ROOT . '/modules/shared/php/core-hacks.php';
 
-// Modules
+// Modules - Telemetry
+if ( class_exists( '\Automattic\VIP\Telemetry\Tracks' ) ) {
```
ingeniumed, line 73: We should do more than this imo. We incorporate the same `is_wpvip_site` check that we have on the block data api. That way, it would only activate this class, and send telemetry data if its an actual VIP site with a valid ID that we can use.

ingeniumed (approved):
Couple of minor changes that are about shifting the tracking calls around and adding more data to be tracked. The core logic is good imo and I'm super pumped to see this go in.

ingeniumed:
I meant to say comments instead of approve but hit approve instead. The EM fields tracking is still left to be added in, that should complete this PR

modules/custom-status/custom-status.php
alecgeatches: Great question! I've been curious about this but hadn't tracked it down yet. After a quick investigation, it appears to be because this is a PHPCS warning, and we ignore those in our `phpcs`/`phpcbf` runs:

https://github.com/Automattic/vip-workflow-plugin/blob/e7e6d82db8b2f21f7fd84dee393a657c443831b7/composer.json#L23-L24

From running `./vendor/bin/phpcs --help`, we can see the `-n` parameter is doing this:

```
  -n        Do not include warnings. Shortcut for "--warning-severity=0".
```

What's happening here is my local PHPCS extensions is auto-fixing this warning, but it's being ignored by our tests, even though it's a warning in our rules. If you're curious, here's the result of running `phpcs` without `-n`:

```
./vendor/bin/phpcs -p -s -v --standard="phpcs.xml.dist" --extensions=php --ignore="/vendor/*,/node_modules/*"

# ...

FILE: modules/shared/php/core-hacks.php
----------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
----------------------------------------------------------------------------------------------------------------------------------
 85 | WARNING | The method parameter $post_parent is never used (Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed)
----------------------------------------------------------------------------------------------------------------------------------


FILE: modules/custom-status/rest/custom-status-endpoint.php
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 25 WARNINGS AFFECTING 25 LINES
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  35 | WARNING | [x] Array double arrow not aligned correctly; expected 18 space(s) between "'name'" and double arrow, but found 14. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
  46 | WARNING | [x] Array double arrow not aligned correctly; expected 11 space(s) between "'description'" and double arrow, but found 7. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
  62 | WARNING | [x] Array double arrow not aligned correctly; expected 5 space(s) between "'required_user_ids'" and double arrow, but found 1.
     |         |     (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
  81 | WARNING | [x] Array double arrow not aligned correctly; expected 18 space(s) between "'name'" and double arrow, but found 14. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
  90 | WARNING | [x] Array double arrow not aligned correctly; expected 20 space(s) between "'id'" and double arrow, but found 16. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 103 | WARNING | [x] Array double arrow not aligned correctly; expected 11 space(s) between "'description'" and double arrow, but found 7. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 119 | WARNING | [x] Array double arrow not aligned correctly; expected 5 space(s) between "'required_user_ids'" and double arrow, but found 1.
     |         |     (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 201 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 18 spaces but found 15 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 202 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 18 spaces but found 15 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 203 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 11 spaces but found 8 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 205 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 5 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 234 | WARNING | [x] Array double arrow not aligned correctly; expected 18 space(s) between "'name'" and double arrow, but found 15. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 235 | WARNING | [x] Array double arrow not aligned correctly; expected 11 space(s) between "'description'" and double arrow, but found 8. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 236 | WARNING | [x] Array double arrow not aligned correctly; expected 18 space(s) between "'slug'" and double arrow, but found 15. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 238 | WARNING | [x] Array double arrow not aligned correctly; expected 5 space(s) between "'required_user_ids'" and double arrow, but found 1.
     |         |     (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 253 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 22 spaces but found 19 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 254 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 18 spaces but found 15 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 255 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 18 spaces but found 15 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 256 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 11 spaces but found 8 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 258 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 5 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 275 | WARNING | [x] Equals sign not aligned correctly; expected 1 space but found 3 spaces (Generic.Formatting.MultipleStatementAlignment.IncorrectWarning)
 300 | WARNING | [x] Array double arrow not aligned correctly; expected 18 space(s) between "'name'" and double arrow, but found 15. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 301 | WARNING | [x] Array double arrow not aligned correctly; expected 11 space(s) between "'description'" and double arrow, but found 8. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 302 | WARNING | [x] Array double arrow not aligned correctly; expected 18 space(s) between "'slug'" and double arrow, but found 15. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 304 | WARNING | [x] Array double arrow not aligned correctly; expected 5 space(s) between "'required_user_ids'" and double arrow, but found 1.
     |         |     (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 25 MARKED SNIFF VIOLATIONS AUTOMATICALLY
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/shared/php/util.php
--------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 6 WARNINGS AFFECTING 3 LINES
--------------------------------------------------------------------------------------------------------------------------------------------------------
 60 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 60 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 61 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 61 | WARNING | Detected usage of a non-sanitized input variable: $_GET['preview_id'] (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
 62 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 62 | WARNING | Detected usage of a non-sanitized input variable: $_GET['preview_nonce'] (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
--------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/shared/php/options-utilities.php
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 4 WARNINGS AFFECTING 4 LINES
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 14 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 6 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 26 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 5 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 80 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 8 spaces but found 7 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 81 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 8 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 4 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/settings/settings.php
---------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------------------
 125 | WARNING | The method parameter $args is never used (Generic.CodeAnalysis.UnusedFunctionParameter.Found)
---------------------------------------------------------------------------------------------------------------


FILE: modules/shared/php/helper-utilities.php
---------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 7 WARNINGS AFFECTING 6 LINES
---------------------------------------------------------------------------------------------------------------------
  68 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
  68 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
  76 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
  77 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
  82 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 100 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 105 | WARNING | Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
---------------------------------------------------------------------------------------------------------------------


FILE: modules/settings/views/settings.php
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 26 | WARNING | [x] Array double arrow not aligned correctly; expected 15 space(s) between "'type'" and double arrow, but found 1. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 27 | WARNING | [x] Array double arrow not aligned correctly; expected 8 space(s) between "'dismissible'" and double arrow, but found 1. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/shared/php/meta-cleanup-utilities.php
------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
------------------------------------------------------------------------------------------------------------------------------------------------------------
 26 | WARNING | [x] Equals sign not aligned correctly; expected 1 space but found 2 spaces (Generic.Formatting.MultipleStatementAlignment.IncorrectWarning)
------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/preview/preview.php
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 47 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 1 space but found 2 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 48 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 3 spaces but found 4 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/editorial-metadata/editorial-metadata.php
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 11 WARNINGS AFFECTING 11 LINES
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 283 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 32 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 284 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 5 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 307 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 4 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 320 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 9 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 321 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 1 space but found 2 spaces (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 326 | WARNING | [x] Usage of ELSE IF is discouraged; use ELSEIF instead (PSR2.ControlStructures.ElseIfDeclaration.NotAllowed)
 335 | WARNING | [x] Usage of ELSE IF is discouraged; use ELSEIF instead (PSR2.ControlStructures.ElseIfDeclaration.NotAllowed)
 358 | WARNING | [x] Usage of ELSE IF is discouraged; use ELSEIF instead (PSR2.ControlStructures.ElseIfDeclaration.NotAllowed)
 366 | WARNING | [x] Array double arrow not aligned correctly; expected 8 space(s) between "'name'" and double arrow, but found 4. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 367 | WARNING | [x] Array double arrow not aligned correctly; expected 8 space(s) between "'slug'" and double arrow, but found 4. (WordPress.Arrays.MultipleStatementAlignment.DoubleArrowNotAligned)
 394 | WARNING | [x] Usage of ELSE IF is discouraged; use ELSEIF instead (PSR2.ControlStructures.ElseIfDeclaration.NotAllowed)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 11 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/notifications/notifications.php
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 1 WARNING AFFECTING 1 LINE
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 257 | WARNING | `wp_mail` should be used sparingly. For any bulk emailing should be handled by a 3rd party service, in order to prevent domain or IP addresses being flagged as spam.
     |         | (WordPressVIPMinimum.Functions.RestrictedFunctions.wp_mail_wp_mail)
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


FILE: modules/custom-status/custom-status.php
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 0 ERRORS AND 5 WARNINGS AFFECTING 3 LINES
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  35 | WARNING | [x] Equals sign not aligned with surrounding assignments; expected 10 spaces but found 1 space (Generic.Formatting.MultipleStatementAlignment.NotSameWarning)
 383 | WARNING | [ ] json_encode() is discouraged. Use wp_json_encode() instead. (WordPress.WP.AlternativeFunctions.json_encode_json_encode)
 922 | WARNING | [ ] Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 922 | WARNING | [ ] Processing form data without nonce verification. (WordPress.Security.NonceVerification.Recommended)
 922 | WARNING | [ ] Detected usage of a non-sanitized input variable: $_REQUEST['post_status'] (WordPress.Security.ValidatedSanitizedInput.InputNotSanitized)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Time: 660ms; Memory: 14MB
```

alecgeatches: It looks like most of these are alignment, with a few other changes like `else if` -> `elseif` and some nonce checks we'll either need to add or ignore.

Since we don't have a lot of these, I think it would make sense to remove `-n` in another PR along with the fixes. This would help avoid inconsistencies with local formatting tools, and I think our code is in a close enough state to just remove these last warnings. What do you think?

modules/custom-status/custom-status.php
alecgeatches: Good thinking! I changed this in https://github.com/Automattic/vip-workflow-plugin/pull/21/commits/8c42cb3d0ed78ee3590f0b3595431082614603b6 to go one step further and wait until after `get_custom_status_by()` is called, and pass that into the action.

On the receiving side, we now use those meta fields to track the count of required users and required EM fields on custom status changes using the method you suggested above.

modules/custom-status/custom-status.php
alecgeatches: See [this comment above](https://github.com/Automattic/vip-workflow-plugin/pull/21#discussion_r1797319485), did exactly that.

modules/notifications/notifications.php
alecgeatches: Fixed in https://github.com/Automattic/vip-workflow-plugin/pull/21/commits/b67ae56837f3e6579a24925549319659f3248208. This one is a bit tricky because `get_options_by_key()` has a return type of `string|array|bool|null`, so we need to be extra explicit here about what type we're expecting.

vip-workflow.php
alecgeatches: I didn't implement this comment (yet) because I think [`is_wpvip_site()`](https://github.com/Automattic/vip-block-data-api/blob/079599c5c30619810d59667b3b600edb206587c1/src/analytics/analytics.php#L99-L104) is actually a bit more restrictive than necessary. The Block Data API implementation has additional checks that we're not sandboxed and running on actual VIP infrastructure before we log data points. However, mu-plugin's `Tracks` library works fine on local installs and just tracks that information with the event (`vipgo_env`).

I think I've found that seeing local datapoints in Tracks is useful for testing, as well as a feature of the mu-plugins library that we don't need to work around. Instead of making our own decisions about the proper environment variables, we're leaving it up to the telemetry library to determine if it's proper to log data, which I think makes more sense here.

alecgeatches:
@ingeniumed I'm going to merge this PR now and finish the EM telemetry in a separate PR.

PR #23, opened 2024-08-27
Add editorial metadata fields to the block editor

ingeniumed:
## Description

This PR is meant to add the Editorial Metadata fields to the block editor, with some tweaks to the way everything works:

- The type cannot be changed for a field. We may add support for this down the line, but for now just the name and description can be changed. I'd like to see how we can allow for migration of data before we add that in.
- We are going to be starting with the text and checkbox types to keep things simple. The date type would have been included by Gutenberg's DateTimePicker doesn't allow for labels/help and looks quite cluttered compared to the other fields. We can add this later on.
- There's no divider between the fields, but personally I think it's not necessary. Open to feedback on this though.
- I've added the type to the cards, and ensured the first letter is capitalized. I've left it as is and not made it strong. Again, open to feedback on this.

Screenshot 2024-08-27 at 3 28 09 PM

Screenshot 2024-08-27 at 2 52 18 PM

Once the PR's design is finalized, I'll update the readme with a screenshot of the admin and block editor as well.


class-workflow.php
```diff
@@ -118,7 +118,7 @@ private function load_modules() {
 	 * @uses add_action() To add various actions
 	 */
 	private function setup_actions() {
-		add_action( 'init', [ $this, 'action_init' ] );
+		add_action( 'init', [ $this, 'action_init' ], 8 );
```
ingeniumed, line 121: Credit to @alecgeatches for helping with this, as I kept running into an inception situation.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -12,17 +12,17 @@
...
 	// ToDo: Review the default metadata types we provide OOB
 	const SUPPORTED_METADATA_TYPES = [
 		'checkbox',
-		'date',
-		'number',
 		'text',
 	];
 	const METADATA_TAXONOMY = 'vw_editorial_meta';
-	const METADATA_POSTMETA_KEY = '_vw_editorial_meta';
+	const METADATA_POSTMETA_KEY = 'vw_editorial_meta';
```
ingeniumed, line 25: It wouldn't be picked up in gutenberg with that _ at the start

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -264,6 +302,8 @@ public function insert_editorial_metadata_term( $args ) {
 		if ( is_wp_error( $inserted_term ) ) {
 			return $inserted_term;
 		} else {
+			// Update the term with the meta_key, as we use the term_id to generate it
+			$this->update_editorial_metadata_term( $inserted_term['term_id'] );
```
ingeniumed, line 314: Storing the meta key with the term to make it easier to keep track of the post meta keys. The term_id isn't available until after we save it, so this weird update needs to be done but it's fast enough that it's not a problem

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -332,6 +375,10 @@ public function update_editorial_metadata_term( $term_id, $args ) {
 	 * @return bool $result Whether or not the term was deleted
 	 */
 	public function delete_editorial_metadata_term( $term_id ) {
+		$term = $this->get_editorial_metadata_term_by( 'id', $term_id );
+		$post_meta_key = $this->get_postmeta_key( $term_id, $term->type );
+		delete_post_meta_by_key( $post_meta_key );
```
ingeniumed, line 388: This ensures all the keys are gone, before we delete the term. Not checking the result because if a post doesn't use it it will give back false. That's expected technically

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -344,6 +391,57 @@ public function delete_editorial_metadata_term( $term_id ) {
...
+				switch ( $arg_type ) {
+					case 'boolean':
+						return boolval( $value );
+					case 'string':
+						return stripslashes( wp_filter_nohtml_kses( trim( $value ) ) );
+				}
+			},
+			'auth_callback' => function () {
+				// ToDo: Review the permissions required to edit metadata
+				return current_user_can( 'edit_posts' ) || current_user_can( 'edit_pages' );
```
ingeniumed: Does this make sense?

README.md
```diff
@@ -109,6 +112,16 @@ Anybody with a preview link (including not logged-in users) will be able to view
 2. Via one-time usage. If the "One-time use" checkbox is selected, the preview URL will only work a single time, and then the token will become invalid. If this box is not selected, a preview URL can be used any number of times before expiration.
 3. Via post status changes. If a post moves out of VIP Workflow's set of extended post statuses, tokens will no longer be valid. For example, a post moved to "Publish" will no longer have valid preview tokens.
 
+#### Editorial Metadata
+
+VIP Workflow adds a "Editorial Metadata" section to the post sidebar, which allows for additional data to be included with the post such as "Needs Legal Review". This can be managed under the plugin's settings, to get a visual for all of the configured editorial metadata fields.
```
alecgeatches, line 117: Can we add a screenshot of the sidebar's Editorial Metadata section? It looks good.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -46,10 +46,36 @@ public function init() {
...
 		$this->register_taxonomy();
 
+		EditEditorialMetadata::init();
+
 		// Load CSS and JS resources that we probably need in the admin page
 		add_action( 'admin_enqueue_scripts', [ $this, 'action_admin_enqueue_scripts' ] );
 
-		EditEditorialMetadata::init();
+		// Load block editor JS
+		add_action( 'enqueue_block_editor_assets', [ $this, 'load_scripts_for_block_editor' ], 9 /* Load before custom status module */ );
```
alecgeatches: We *should* be able to use script dependencies rather than explicit priorities to be able to decide load order. In `load_scripts_for_block_editor`, maybe this would work:

```php
$asset_file = include VIP_WORKFLOW_ROOT . '/dist/modules/editorial-metadata/editorial-metadata-block.asset.php';
$dependencies = array_merge( $asset_file['dependencies'], [ 'vip-workflow-block-custom-status-script' ] );

wp_enqueue_script( 'vip-workflow-block-editorial-metadata-script', VIP_WORKFLOW_URL . 'dist/modules/editorial-metadata/editorial-metadata-block.js', $dependencies, $asset_file['version'], true );
```

I see this is also being done in the preview module (probably my code), and that could potentially be fixed as well.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -46,10 +46,36 @@ public function init() {
...
+	 * Register the post meta for each editorial metadata term
+	 */
+	public function register_editorial_metadata_terms_as_post_meta() {
+		$editorial_metadata_terms = $this->get_editorial_metadata_terms();
+
+		foreach ( $editorial_metadata_terms as $term ) {
+			$post_meta_key = $this->get_postmeta_key( $term );
+			$post_meta_args = $this->get_postmeta_args( $term );
+
+			// ToDo: Fix the bug where if a post type is later unsupported, the editorial metadata fields still show up for that post type
```
alecgeatches: Can you explain this issue? I'd think if a post type was unsupported, it wouldn't show up in `get_supported_post_types()` and then wouldn't get registered.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -136,6 +156,20 @@ public function action_admin_enqueue_scripts() {
 		}
 	}
 
+	public function load_scripts_for_block_editor() {
+		$asset_file = include VIP_WORKFLOW_ROOT . '/dist/modules/editorial-metadata/editorial-metadata-block.asset.php';
+		wp_enqueue_script( 'vip-workflow-block-editorial-metadata-script', VIP_WORKFLOW_URL . 'dist/modules/editorial-metadata/editorial-metadata-block.js', $asset_file['dependencies'], $asset_file['version'], true );
+
+		$editorial_metadata_terms = $this->get_editorial_metadata_terms();
+		wp_localize_script( 'vip-workflow-block-editorial-metadata-script', 'VipWorkflowEditorialMetadatas', $editorial_metadata_terms );
```
alecgeatches: I've noticed we use a variety of styles for the localized global (e.g. `VW_EDITORIAL_METADATA_CONFIGURE` or `VipWorkflowEditorialMetadatas`), and also sometimes passing an array directly or an array of values.

I'd personally prefer we use the caps-underscore style, and pass any values as a named key, like:


```suggestion
		wp_localize_script( 'vip-workflow-block-editorial-metadata-script', 'VW_EDITORIAL_METADATA', [
		    'terms' => $editorial_metadata_terms
		] );
```

The named-key style just makes it easier to pass another bit of data into the page without affecting how the terms are being consumed.

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Due to the way the `register_post_meta` operation works, if a post meta has been registered for say a `page` and then later support for a `page` is removed then that post_meta would still show up for a page. I'm leaning towards checking the supported post type in the block editor to manually stop the editorial metadata from loading.

README.md
ingeniumed: Yip, I mentioned in the PR that I'd add that after the PR was finalized just to ensure only the final design gets uploaded to the media branch

modules/editorial-metadata/editorial-metadata.php
ingeniumed: I'll give this a shot

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Agreed on  that, I'll make that change

modules/editorial-metadata/editorial-metadata.php
alecgeatches: I wasn't sure how this worked, so I spent some time in the debugger. The short answer is I think you can remove `auth_callback` entirely as core handles permission checks properly unless you want to check extra permissions.

When editorial metadata is updated, this initially travels through `update_meta_value()` [to the `current_user_can()`](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/rest-api/fields/class-wp-rest-meta-fields.php#L390) function. This eventually falls to [`WP_User::has_cap()`](https://github.com/WordPress/WordPress/blob/741a7869fd77a0f979a2116ae87b1413a0e62fb1/wp-includes/class-wp-user.php#L776) which checks the `"edit_post_meta"` capability with the current post ID. When the post is a page, this is `map_meta_cap()`ed into the `"edit_pages"` capability, and when it's a post it becomes `"edit_posts"`. If you don't pass a custom auth callback, it seems like WordPress maps to the relevant capability type properly.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -344,6 +393,57 @@ public function delete_editorial_metadata_term( $term_id ) {
...
+				return current_user_can( 'edit_posts' ) || current_user_can( 'edit_pages' );
+			},
+		];
+		return $args;
+	}
+
+	/**
+	 * Generate a unique key based on the term id and type
+	 *
+	 * Key is in the form of _vw_editorial_meta_{type}_{term_id}
```
alecgeatches: Teensy change:

```suggestion
	 * Key is in the form of vw_editorial_meta_{type}_{term_id}
```

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
```diff
@@ -131,6 +131,10 @@ export default function EditorialMetadataManager( {
 											<Text>
 												<i>{ eMetadataTerm.description }</i>
 											</Text>
+											<Text>
+												{ eMetadataTerm.type.charAt( 0 ).toUpperCase() +
```
alecgeatches: This took me a bit to figure out. Could you maybe assign this to a variable like `metadataTypeUppercased` and put that in the `<Text>` instead?

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
```diff
@@ -150,15 +154,17 @@ export default function EditorialMetadataManager( {
...
-												<Button
-													size="compact"
-													variant="primary"
-													icon={ 'edit' }
-													onClick={ () => {
-														setEMetadataTerm( eMetadataTerm );
-														setIsCreateEditModalVisible( true );
-													} }
-												></Button>
+												{
```
alecgeatches: Is this scope doing anything?

modules/editorial-metadata/lib/editorial-metadata-block.js
```diff
@@ -0,0 +1,85 @@
...
+	term_id: editorialMetadata.term_id,
+	description: editorialMetadata.description,
+} ) );
+
+const noEditorialMetadatasToShow = editorialMetadatas.length === 0;
+
+const CustomMetaPanel = ( { metaFields, setMetaFields } ) => (
+	<PluginDocumentSettingPanel name="editorialMetadataPanel" title="Editorial Metadata">
+		{ noEditorialMetadatasToShow && (
+			<Text>Configure your editorial metadata, within the VIP Workflow Plugin Settings.</Text>
```
alecgeatches: Should internationalize this text.

alecgeatches (approved):
Left some minor comments above. Overall, this looks and works really well!

I still am noticing some odd layout behavior in the configuration page. For example, this was after adding two new metadata options to the default set:

Screenshot 2024-08-28 at 3 26 16 PM

and then after adding two more:

Screenshot 2024-08-28 at 3 26 57 PM

This isn't a blocker for this PR, but they do seem wonky. I think we should fix this with a different layout pattern, probably.

README.md
alecgeatches: My bad, forgot about that! 👍

alecgeatches:
One more comment, I do think order is important for these metadata. If I want to add a new "Super Important Checkbox For Legal Reasons" to the top of my editorial metadata, it seems like the only way is to delete all existing metadata, add that one, and then add them back. Or if I have a "Final Approval" checkbox I want to show last to follow the process chronologically, it will only work if it's the last metadata I added.

Because editors will be interacting with these values in their process, ordering will matter to them. As above, not a blocker for this PR, but something I think we need in the future.

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Beauty, I struggled a bit to dig into that exact `map_meta_cap` call so appreciate you digging that up! I'll take this out 👍🏾

modules/editorial-metadata/editorial-metadata.php
ingeniumed: Based on my testing, this works! I'll make the change in the preview module as well

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
ingeniumed: Yip, can do

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
ingeniumed: Nope, I had been doing some testing with layouts and I must have not reverted this

modules/editorial-metadata/lib/components/editorial-metadata-manager.js
ingeniumed: Moved this to a function, and I fixed the discrepancy between what's shown here and in the type dropdown when creating or editing an editorial metadata field.

ingeniumed:
I'm def getting what you mean. I'm thinking that we re-use the custom status code so that way it gets all that functionality. It would be a vertical list in that case. This would be something that I can take care of in the design review PR. Perhaps I could clean up the code a bit as a result.

I'll also add a screenshot in that PR.

modules/editorial-metadata/editorial-metadata.php
ingeniumed: This won't work for the preview module, it results in a blank block editor. I tried to dig into it and fix it to no avail so I've skipped making this change for the preview module. I'm guessing this is because the preview module depends on the custom status module so it gets into some kind of a weird loop. If EM ever depends on that, we might have the same issue

README.md
ingeniumed: Given the design change discussion that we had below, let's skip this. If we aren't able to do the design review PR in time for the beta then we can make a PR to add this in.

PR #25, opened 2024-08-28
adding content and updating blueprint

smithjw1:
## Description

I added a few pieces of content and tagged them with custom statuses. We should expand this over time to include metadata and workflow rules.

## Steps to Test

This is hard to test without merging since it depends on the content export existing on GitHub. Ideas are welcome.


alecgeatches (approved):
Good idea! I'm happy to just merge this and test it live.

PR #27, opened 2024-08-30
Make the changes from the design review

ingeniumed:
## Description

These changes are coming as a result of a design review conducted for the plugin. There's a lot in here so I'm going to summarize as best as I can with text and screenshots.

### custom status

I have switched the UI to be using floating cards that are white, and don't have a background. The background is hidden as removing it caused problems. I've set `starting point` as the top, and drawn an arrow from that to the first status. I have then drawn another arrow from the final status to the add button. I have also updated the description of the module to provide better instructions. I have switched the success notice to be a standard notice because i realized the previous way looked odd when it came to the overall UI. It's still not the best because honestly it pushes the whole UI down but I also wanted to take away the VIP Workflow title. I have also changed the way this module worked to now be truly rest based - the CRUD operations give back individual items.

There's a bug that's been fixed in here and in editorial metadata which is to correct the position when a delete happens.

#### To Do

- Make the arrow actually touch the text and custom status on top, and the button and custom status on the bottom
- Delete the unnecessary transparent container for the statuses. I can't remove it as it causes everything to fail to load.
- See if the success notice could be switched to a snackbar, I failed to do so.

### Editorial Metadata

I didn't really touch this. I wanted to make this the same as custom status but honestly that's going to be a lot of unnecessary scrolling to do. So I've left it and we can see what the feedback says.

#### To Do

- See if you can fix the weird layout quirks, set aside the lack of re-ordering.

### Readme

I haven't updated the readme yet, so this has to be done for some final screenshots and any necessary details for easy usage of the plugin.

## To Do

- If you can, see if you can strip out some common components between custom status and editorial metadata. This isn't necessary right now, but if it doesn't take too long it's worth doing.
- Update the version of the plugin and cut a new release

Screenshot 2024-08-30 at 4 02 44 PM


alecgeatches:
@ingeniumed I didn't make all of the improvements you mentioned, but I did make some changes:


flow-page

- Limited the horizontal width of individual cards and widened the space that a status description can use.
- Removed `<WorkflowArrow>` entirely in favor of a CSS-based solution that better matches our design goals.
- Put the "Publish" status text back at the end, because that's still how our workflow works and it makes the timeline make more sense for now. Also moved the button before the "Publish" text for continuity.
- Removed the successful reorder message. Having that pop-in during a reorder is disorienting and I think unnecessary. Errors will still show an interruptive error message at the top of the page, but for now successful reorders simply do nothing on success. I do think a snackbar component could suffice here, but I did not do that today.

---

> ### Editorial Metadata
> - See if you can fix the weird layout quirks, set aside the lack of re-ordering.

After a bit of playing around with this I think the best solution is probably switching to a CSS grid. I did not do that, because a grid doesn't necessarily lend itself to a reordering UI either, and we're probably going to want to change that when we add support for reordering anyway.

---

> ## To Do
> If you can, see if you can strip out some common components between custom status and editorial metadata. This isn't necessary right now, but if it doesn't take too long it's worth doing.

Didn't want to do a change like this right before release, but definitely a good target for refactoring soon.

---

I'll update the README and create a new release shortly.

PR #28, opened 2024-09-03
Update README.md

smithjw1:
## Description

Making a few changes to the README pre-beta.

## Steps to Test

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,9 @@ Note that, these statuses are also available in the quick edit experience on the
 
 The plugin doesn't expect any specific configuration, so your first step is to set up statuses that reflect your workflow. You may notice that the steps are listed in a linear order. The plugin assumes a linear workflow where content is moving from creation to publish.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,7 +132,7 @@ VIP Workflow adds a "Editorial Metadata" section to the post sidebar, which allo
 
 ### Editorial Metadata
 
-A post type that's supported at first, and then unsupported is used to make posts, then those posts will still show editorial metadata fields.
+If a post is created as a type that supports Editorial Metadata, that metadata will remain even if the post type is modified not to support Editorial Metadata. In other words, once a post is created with Editorial Metadata fields, those fields will remain regardless of configuration changes.
```
smithjw1, line 137: @ingeniumed I'm not sure if I have this right, but I thought I would take a stab.

smithjw1: Also, we don't include any information about what post types will have editorial metadata or how to change that. I would guess it is the same as the post types that include custom statuses, but that isn't explicit.

ingeniumed: This is great, I like the way that it's been phrased

ingeniumed (approved):

PR #30, opened 2024-09-11
Add typescript support

ingeniumed:
## Description

The purpose of this PR is to do the following:

- Lay the groundwork for typescript support. This doesn't do any changes to any existing files because we are re-working the module management system. I'd rather let that finish first before we start to touch existing files. So this just ensures we have the pieces in place to take advantage of it.
- Add stylelint to ensure that we lint our CSS files. That's something that we aren't doing before.
- Update our linting code to run properly

Credit to the remote-data-blocks plugin for a lot of the code that I've adopted.


.github/workflows/php-tests.yml
```diff
@@ -44,13 +44,11 @@ jobs:
...
           composer-options: --prefer-dist --no-progress
 
-      - name: Run PHPCS
-        run: composer run cs
-
       - name: Setup wp-env
         run: wp-env start
         env:
           WP_ENV_CORE: WordPress/WordPress#${{ matrix.wp }}
+          WP_ENV_PHP_VERSION: ${{ matrix.php }}
```
alecgeatches, line 55: 👍

modules/custom-status/lib/configure.scss
```diff
@@ -2,21 +2,29 @@
 	display: flex;
 }
 
-// Draw background line connecting statuses, add overlay background to areas where the line should be hidden
+.status-start,
```
alecgeatches, line 5: Why did these rules change order in this file?

alecgeatches (approved):
I had one question about a CSS file reordering, but nothing important. This looks great! I also appreciate the CSS file linting and cleanup.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
+
+      - name: Setup Node.js
+        uses: Automattic/vip-actions/nodejs-setup@trunk
+        with:
+          node-version-file: .nvmrc
```
chriszarate: Can you use the VIP action for this? Less boilerplate and implements caching that saves time and $$

https://github.com/Automattic/vip-actions/blob/trunk/nodejs-setup/README.md

chriszarate: I don't think you need to setup PHP for this job btw

.github/workflows/lint.yml
```diff
@@ -0,0 +1,68 @@
...
```
chriszarate: Node.js needed?

tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "noImplicitAny": true,
+    "noUncheckedIndexedAccess": true,
+    "outDir": "./dist/",
+    "paths": {
+      "@/*": [ "./modules/*" ]
+    },
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "es5",
```
chriszarate, line 18: Does this match some support target? In all likelihood you can go much higher here, ES5 / ES2009 has been universally supported for more than a decade.

jsconfig.json
```diff
@@ -1,10 +1,9 @@
-
 {
-	"compilerOptions": {
-		"module": "ES6",
-		"jsx": "preserve",
-		"checkJs": true,
-		"allowUnreachableCode": true
-	},
-	"exclude": [ "node_modules" ]
+  "compilerOptions": {
```
chriszarate: is this file still relevant?

alecgeatches: True! This is a stand-in for `tsconfig` [for VSCode](https://code.visualstudio.com/docs/languages/jsconfig), but now we have a legitimate TypeScript configuration.

modules/custom-status/lib/configure.scss
ingeniumed: It's because it was trying to enforce that, if you have a CSS rule targeting one element then do that first followed by targeting any element inside it. So if you notice, `.add-status` is targeted by itself and then within `.status-container` as well. So it's saying, do the outer one first and then the inner one. It's the same thing for the other rules as well.

From what I read, it's supposed to also make you re-think those CSS rules. So for us, we do have two different containers surrounding the draggable statuses that would be good to simplify. But that's been a bit buggy so we have left it for now, would be good to revisit that at some point so it'd be just one container that we have. Then this CSS would be simpler.

Hope all that made sense

.github/workflows/lint.yml
ingeniumed: Good point, I had run into some problems and so I had re-added the checkout code. It can be removed since the vip-action already takes care of that. I'll take out the PHP code as well

.github/workflows/lint.yml
ingeniumed: Since I moved the linting code for PHP from `composer.json` to `package.json` it's needed. Just felt nice having everything in one spot to use

tsconfig.json
ingeniumed: This would be a copy paste from gutenberg/remote-data-blocks that I didn't alter until we actually migrated a JS file to TS. I'd agree with you in that, we can definitely go a lot higher but just left that for a follow up migration PR.

jsconfig.json
ingeniumed: It's def not relevant still, can be killed off

ingeniumed: Done

.github/workflows/lint.yml
ingeniumed: Done

.github/workflows/lint.yml
ingeniumed: Done

ingeniumed:
I've also added husky and lint-staged so that when committing we automatically format all the files. This is something we have in the governance plugin and will ensure we don't miss out on any linting. We do have a check obviously when the PR is made but this will automate that.

PR #31, opened 2024-09-11
Refactor Preview module as static module

alecgeatches:
## Description

The original plan was to remove the module system from the entire codebase. I've since scaled this back to one of the newer modules (Preview) to demonstrate the basics. We can build on this system with other modules and remove the entire module system loader once the other modules has been migrated.

## Steps to Test

1. Check out PR.
2. Ensure generating and previewing links continue to work as designed.

## Other notes

I'll leave some notes in code below describing the basic structure a little better. This isn't directly documented in code, as the structure will be more easy to follow and self-evident once we've moved additional modules to match.

### DIFF
--- a/class-workflow.php
+++ b/class-workflow.php
@@ -75,12 +75,14 @@ class VIP_Workflow {
 		// VIP Workflow base module
 		require_once VIP_WORKFLOW_ROOT . '/modules/shared/php/class-module.php';
 
+		$skip_module_dirs = [ 'shared', 'preview' ];
+
 		// Scan the modules directory and include any modules that exist there
 		$module_dirs = scandir( VIP_WORKFLOW_ROOT . '/modules/' );
 		$class_names = [];
 		foreach ( $module_dirs as $module_dir ) {
 			// Skip the . and .. directories, as well as the shared folder
-			if ( file_exists( VIP_WORKFLOW_ROOT . "/modules/{$module_dir}/$module_dir.php" ) && 'shared' !== $module_dir ) {
+			if ( file_exists( VIP_WORKFLOW_ROOT . "/modules/{$module_dir}/$module_dir.php" ) && ! in_array( $module_dir, $skip_module_dirs ) ) {
 				include_once VIP_WORKFLOW_ROOT . "/modules/{$module_dir}/$module_dir.php";
 
 				// Prepare the class name because it should be standardized
--- a/modules/preview/preview.php
+++ b/modules/preview/preview.php
@@ -130,7 +130,12 @@ class Preview {
 	// Public API
 
 	/**
-	 * Get the expiration options available in the preview modal dropdown.
+	 * Returns the valid set of expiration options for preview links. See the
+	 * vw_preview_expiration_options filter for customization.
+	 *
+	 * @access public
+	 *
+	 * @return array
 	 */
 	public static function get_link_expiration_options(): array {
 		/**
--- a/modules/preview/preview.php
+++ b/modules/preview/preview.php
@@ -12,42 +12,30 @@ require_once __DIR__ . '/rest/preview-endpoint.php';
 
 use VIPWorkflow\Modules\Preview\PreviewEndpoint;
 use VIPWorkflow\VIP_Workflow;
-use VIPWorkflow\Modules\Shared\PHP\Module;
 use VIPWorkflow\Modules\Preview\Token;
+use WP_Query;
 
-class Preview extends Module {
-	public $module;
-
-	public function __construct() {
-		// Register the module with VIP Workflow
-		$this->module_url = $this->get_module_url( __FILE__ );
-
-		$this->module = VIP_Workflow::instance()->register_module( 'preview', [
-			'title'      => __( 'Preview Link', 'vip-workflow' ),
-			'module_url' => $this->module_url,
-			'slug'       => 'preview',
-			'autoload'   => true,
-		] );
-	}
-
+class Preview {
 	/**
 	 * Initialize preview module
 	 */
-	public function init() {
+	public static function init(): void {
 		PreviewEndpoint::init();
 
 		// Load block editor JS
-		add_action( 'enqueue_block_editor_assets', [ $this, 'load_block_editor_scripts' ], 9 /* Load before custom status module */ );
+		add_action( 'enqueue_block_editor_assets', [ __CLASS__, 'load_block_editor_scripts' ], 9 /* Load before custom status module */ );
 
 		// Load block editor CSS
-		add_action( 'enqueue_block_editor_assets', [ $this, 'load_block_editor_styles' ] );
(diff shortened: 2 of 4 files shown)


modules/preview/preview.php
```diff
@@ -12,42 +12,30 @@
...
-
-		$this->module = VIP_Workflow::instance()->register_module( 'preview', [
-			'title'      => __( 'Preview Link', 'vip-workflow' ),
-			'module_url' => $this->module_url,
-			'slug'       => 'preview',
-			'autoload'   => true,
-		] );
-	}
-
+class Preview {
```
alecgeatches, line 18: For now, the `Preview` class does not extend `Module`. The base class adds what are essentially utility methods once the module loading system is removed. The preview module doesn't use any of those, and I think these could be relocated and called directly as a utility methods in other classes that do make use of them.

I'll wait until we migrate something that uses all of these (like `Custom_Status`) to make a judgement call there.

modules/preview/preview.php
```diff
@@ -136,10 +127,12 @@ public function allow_preview_result( $posts, $query ) {
 		return $posts;
 	}
 
+	// Public API
```
alecgeatches, line 130: Because we're now interacting with modules directly (`Preview::get_link_expiration_options()`), it's good to separate `public` functions that are filters (not part of the public API), and `public` functions that are meant to be called from outside modules. I like to put these at the bottom of the class, because `init()` and `init()`-related filters make more sense to me at the top of the class.

modules/preview/preview.php
```diff
@@ -169,3 +162,5 @@ public function get_link_expiration_options() {
 		]);
 	}
 }
+
+Preview::init();
```
alecgeatches, line 171: Like REST API endpoints we use already, the new modules `init()` themselves. Load order shouldn't be a huge problem for a couple of reasons:

- `vip-workflow.php` file determines which order files load. If we need to include a base class or something before other files, we can do it there.
- Otherwise, the `init()` function serves as an entry point to set filters on the WordPress lifecycle. These can use priorities, dependencies, and other WordPress concepts to manage load order. It's worth noting that `init()` is now called before any WordPress filters (as part of initial plugin loading), so we can now set a `'init'` action hook or any other hook at any priority without worrying about [the lower priority trick](https://github.com/Automattic/vip-workflow-plugin/pull/23#discussion_r1732150604) we used before.

Overall, this system has some nice properties. Each module is independent, and in charge of setting and removing their own hooks. Class intercommunication happens explicitly and functionally by calling static methods. There's no magic, and module logic is self-contained.

modules/preview/rest/preview-endpoint.php
```diff
@@ -46,7 +47,7 @@ public static function register_routes() {
 				'expiration'      => [
 					'required'          => true,
 					'validate_callback' => function ( $param ) {
-						$expiration_options = VIP_Workflow::instance()->preview->get_link_expiration_options();
+						$expiration_options = Preview::get_link_expiration_options();
```
alecgeatches, line 50: This is what inter-module communication should look like moving forward, with proper intellisense and no dynamic properties.

vip-workflow.php
```diff
@@ -16,28 +16,34 @@
...
+define( 'VIP_WORKFLOW_ROOT', __DIR__ );
+define( 'VIP_WORKFLOW_URL', plugins_url( '/', __FILE__ ) );
+define( 'VIP_WORKFLOW_SETTINGS_PAGE', add_query_arg( 'page', 'vw-settings', get_admin_url( null, 'admin.php' ) ) );
+define( 'VIP_WORKFLOW_REST_NAMESPACE', 'vip-workflow/v1' );
+
+// Main plugin class
+require_once VIP_WORKFLOW_ROOT . '/class-workflow.php';
+
+// Modules
+require_once VIP_WORKFLOW_ROOT . '/modules/preview/preview.php';
```
alecgeatches, line 49: I've moved all of the base requires to this file, rather than `class-workflow.php`. We really just need a centralized place to include these top-level imports and ensure load order, and this is a simpler and much less busy file than `class-workflow.php`.

Classes that are particular to a module (e.g. `PreviewEndpoint`) are required by the module that uses them.

class-workflow.php
```diff
@@ -75,12 +75,14 @@ private function load_modules() {
 		// VIP Workflow base module
 		require_once VIP_WORKFLOW_ROOT . '/modules/shared/php/class-module.php';
 
+		$skip_module_dirs = [ 'shared', 'preview' ];
```
ingeniumed, line 78: Minor: Add a todo here that this needs to be removed once the re-write is finalized.

modules/preview/preview.php
ingeniumed: Makes sense to me, especially since this module does not take advantage of any options.

modules/preview/preview.php
```diff
@@ -65,25 +53,28 @@ public function load_block_editor_scripts() {
...
+
+	public static function add_preview_query_vars( array $query_vars ) {
+		// Allow 'vw-token' GET parameter to be detected in WP_Query parameters
 		$query_vars[] = 'vw-token';
 
 		return $query_vars;
 	}
 
-	public function allow_preview_result( $posts, $query ) {
+	public static function allow_preview_result( array $posts, WP_Query $query ) {
```
ingeniumed, line 77: Love the types that are being slowly added

modules/preview/preview.php
ingeniumed: We could add the access identifier in the docs to signify how it's to be used, but we could that in a documentation pass later

vip-workflow.php
ingeniumed: Agreed on this. This would be similar to how our other plugins work, wherein the requires happen here but there's still one main entry point for the plugin itself.

ingeniumed (approved):
Love this re-write, it's a lot better now imo.

What I do want to in a follow up PR, after the other modules are migrated is:

- Standardize the folders in each module: rest, block editor, and core feature class
- Move the settings JS to the settings module itself

It'd be nice to get rid of the lib folder structure that's there as it's quite confusing imo as to what it means.

class-workflow.php
ingeniumed: Actually, this is only going to increase with each migration until at the very end when it's all going to be removed so ignore this.

ingeniumed:
This is good to merge in!

modules/preview/preview.php
alecgeatches: That's a good idea. I added a full PHPDoc header and `@access public` in this commit: https://github.com/Automattic/vip-workflow-plugin/pull/31/commits/81074a3fef0313f0f58befc01c366d917219ca76. We may also want to explicitly set a non-public access for other methods we use, like filter callbacks, but I haven't implemented that for this PR.

alecgeatches:
@ingeniumed Thank you for the review! Merging.

PR #32, opened 2024-09-12
Add a notice when the min WP version is not met

ingeniumed:
## Description

Same PR as [this](https://github.com/Automattic/vip-governance-plugin/pull/86).

It'll enforce the min WP version.


.eslintignore
```diff
@@ -2,3 +2,4 @@
 node_modules
 vendor
 *.php
+dist
```
ingeniumed, line 5: Missed this in the typescript PR so adding it in

PR #34, opened 2024-09-16
Refactor editorial metadata module

alecgeatches:
## Description

Following https://github.com/Automattic/vip-workflow-plugin/pull/31, this refactors the editorial metadata module to a static functional class that does not use the original EditFlow module system. These changes are part of an overall refactor to remove use of dynamic properties from `VIP_Instance`, improve typing, and simplify module intercommunication.

## Other notes

- The `VIP_Workflow`-managed system for registering sidebar menus has been simplified so that each module registers itself using the `admin_menu` hook. This changes a couple of things:

    - We no longer rely on the `module-header.php` view for printing the short & long description of a module. Modules have moved that logic into their own views. I did this because we don't necessarily want to have a standardized view for each page, and we don't have a lot of modules that we'd want to keep regulated and consistent like EditFlow does. We may want no header or something else. For now, I just migrated the prior header into each of the module views.
  
    - The system for setting success/error `'messages'` key on reload was only being used for a single success message on the settings page. Given that we're mostly moving to non-refresh reactive UI, I moved this logic into the settings page alone.
- The `install()` function setup that was previously being used in the editorial metadata module is still being used. I think the `install()` and `update()` functions are really nifty parts of EditFlow I'd like to keep. Each module can have a bit of code for managing new or updated data which is great and self-contained. See `InstallUtilities::install_if_first_run()` and the usage of this function in `EditorialMetadata`. This is a less magical and more functional version of the same setup function.
 
I'll add additional notes below in code.  

## Steps to Test

1. Under Admin -> VIP Workflow, review each settings page as they're all being rendered slightly differently.
2. Test the UI on the editorial metadata settings page, CRUD operations, and usage in the block editor.
3. The `install()` function used by `EditorialMetadata` can be tested by:
    1. Go to the Admin -> VIP Workflow -> Editorial Metadata page.

    2. Delete each existing editorial metadata until the page is empty.
    3. Refresh the page. Note that the page is still empty.
    4. In a terminal, delete the `vip_workflow_editorial_metadata_options` option. For `vip dev-env`, do this:
        ```bash
        > vip dev-env shell

        # ...

        $ wp option delete vip_workflow_editorial_metadata_option
        Success: Deleted 'vip_workflow_editorial_metadata_options' option
        ```

    5. Refresh the Editorial Metadata page. The terms provided by `install()` should be re-added to the page.


class-workflow.php
```diff
@@ -372,6 +297,42 @@ public function update_all_module_options( $mod_name, $new_options ) {
...
 		return update_option( $this->options_group . $mod_name . '_options', $this->modules->$mod_name->options );
 	}
+
+
+	/**
+	 * Collect all of the active post types
+	 *
+	 * @return array $post_types All of the post types that are 'on'
+	 */
+	public function get_supported_post_types(): array {
```
alecgeatches, line 307: This function was moved out of the `Module` base class, which just called back into `VIP_Workflow` anyway.

This function relies on some filter timings related to older module registration (see `$this->settings->module->options->post_types` below) that would be tricky to refactor now. When we have moved all modules out of the old system, we can move this into a simpler static settings function.

A lot of code uses `get_supported_post_types()`, so this caused a lot of changes.

class-workflow.php
```diff
@@ -372,6 +297,42 @@ public function update_all_module_options( $mod_name, $new_options ) {
...
+
+		return $post_types;
+	}
+
+	/**
+	 * Whether or not the current page is our settings view. Determination is based on $pagenow, $_GET['page'], and if it's settings module or not.
+	 *
+	 * @return bool $is_settings_view Return true if it is
+	 */
+	public static function is_settings_view_loaded( string $slug ): bool {
```
alecgeatches, line 327: This was also moved from `Module` and simplified. We no longer check if the page matches any module slug, as this caused all VIP Workflow dependencies to be loaded on all VIP Workflow pages. Instead, we pass the current module `$slug` and directly check if the current page is loaded.

modules/custom-status/custom-status.php
```diff
@@ -24,6 +25,8 @@ class Custom_Status extends Module {
 	// This is taxonomy name used to store all our custom statuses
 	const TAXONOMY_KEY = 'post_status';
 
+	const SETTINGS_SLUG = 'vw-custom-status';
```
alecgeatches, line 28: Every module in the new system now has a class-defined `SETTINGS_SLUG` that is used both for the settings page slug and saved module metadata (like `loaded_once`).

modules/custom-status/custom-status.php
```diff
@@ -466,7 +478,7 @@ public function add_custom_status( $term, $args = [] ) {
 			$args['position'] = $default_position;
 		}
 
-		$encoded_description = $this->get_encoded_description( $args );
+		$encoded_description = TaxonomyUtilities::get_encoded_description( $args );
```
alecgeatches, line 481: `TaxonomyUtilities` only currently handles `get_encoded_description()` and `get_unencoded_description()`, but this allowed moving some repeated code out of the `CustomStatus` and `EditorialMetadata` classes.

modules/editorial-metadata/rest/editorial-metadata-endpoint.php
```diff
@@ -6,15 +6,15 @@
...
+use VIPWorkflow\Modules\EditorialMetadata;
 use VIPWorkflow\VIP_Workflow;
 use WP_Error;
 use WP_REST_Request;
 use WP_Term;
 
 defined( 'ABSPATH' ) || exit;
 
-class EditEditorialMetadata {
+class EditorialMetadataEndpoint {
```
alecgeatches, line 17: REST endpoints have been standardized to end with `Endpoint`. I just like this because it's easier to tell the difference between filenames in the editor and gives REST files an easy pattern.

modules/editorial-metadata/editorial-metadata.php
```diff
@@ -1,20 +1,21 @@
 <?php
 
 /**
- * Class Editorial_Metadata
+ * Class EditorialMetadata
```
alecgeatches, line 4: We're using both `Class_Name` and `ClassName` in our code. I like the latter, so slowly converting files to use that version.

modules/settings/settings.php
```diff
@@ -249,7 +252,7 @@ public function settings_validate( $new_options ) {
 	 */
 	public function helper_settings_validate_and_save() {
 
-		if ( ! isset( $_POST['action'], $_POST['_wpnonce'], $_POST['option_page'], $_POST['_wp_http_referer'], $_POST['vip_workflow_module_name'], $_POST['submit'] ) || ! is_admin() ) {
+		if ( ! isset( $_POST['action'], $_POST['_wpnonce'], $_POST['option_page'], $_POST['_wp_http_referer'], $_POST['submit'] ) || ! is_admin() ) {
```
alecgeatches, line 255: Removed `vip_workflow_module_name` from tests because we only have a single page that submits settings now.

modules/shared/php/options-utilities.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @desc Utility methods for initializing a module after first install
+ */
+
+namespace VIPWorkflow\Modules\Shared\PHP;
+
+use stdClass;
+use VIPWorkflow\VIP_Workflow;
+
+class OptionsUtilities {
```
alecgeatches, line 13: This is a set of functional utilities to manage options like `loaded_once` that's a bit less magic than before.

class-workflow.php
ingeniumed: Could you add this as a ToDo? Trying to make sure everything like this is noted as a todo, so we can get to it eventually.

ingeniumed (approved):
Just the one minor comment otherwise this is awesome. 

Nice work, can't wait to do the same to the custom status module and simplify the whole thing!

alecgeatches:
@ingeniumed Thank you for the review! Added your suggested ToDo comment in https://github.com/Automattic/vip-workflow-plugin/pull/34/commits/e99a1ef82de055e6207080fd0e587bbcec069aa1. Going to merge!

PR #35, opened 2024-09-17
Change email notifications to not go to blog admins

ingeniumed:
## Description

This tweaks the notifications system to be off by default, until an email address and/or webhook notification is configured. This way, blog admins are not always notified as that doesn't quite make sense. The recipients can be modified via filters, until we support this via the custom status.

I have also modified the whole settings configuration to show two textboxes that can be tweaked, rather than turning it on/off and then configuring it. That feels closer to how WordPress does it in the settings page.

Screenshot 2024-09-16 at 3 54 31 PM

In addition, the email body has been tweaked to be a lot simpler. I tried to get it closer to the VIP notifications way of doing it without changing it to an HTML email body. This is what it looks like:

Screenshot 2024-09-17 at 3 35 14 PM

It's simple, and maybe HTML does make sense but figured for how now this would be better. I haven't incorporated a filter in, as we would need to allow changing the subject/body/headers. So for now it's just a simple one.

## Steps to Test

- Ensure mailpit and cron are configured on `dev-env`
- Configure a dummy email and a webhook
- Create a new post, and publish it
- Go to your mailpit url, and ensures the emails show up.


modules/shared/php/views/module-header.php
```diff
@@ -5,6 +5,8 @@
 ?>
 
 <div class="wrap vip-workflow-admin">
+	<?php /* ToDo: Convert this message into a proper notice for errors and success messages */ ?>
```
ingeniumed: Realized this was removed in my previous PR so I've added it back in

modules/settings/settings.php
```diff
@@ -260,10 +222,12 @@ public function helper_settings_validate_and_save() {
 			return false;
 		}
 
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- wp_verify_nonce() is already doing the necessary work
```
ingeniumed: This really shouldn't be needed but given this whole code is going to be simplified, it's fine for now

README.md
```diff
@@ -165,6 +163,32 @@ add_filter( 'vw_notification_ignored_statuses', function ( $ignored_statuses, $p
...
+*/
+apply_filters( 'vw_notification_email_recipients', $email_recipients, $post );
+```
+
+For example, this filter can be used to send email notifications to more than just 1 recipients especially for special statuses:
+
+```php
+add_filter( 'vw_notification_email_recipients', function ( $email_recipients, $action, $post  ) {
+  if ( $post->post_status === 'legal-review' ) {
+    $email_recipients[] = 'saul.goodman@sgoodmanassoc.com';
```
alecgeatches, line 189: Love it

modules/notifications/notifications.php
```diff
@@ -217,19 +193,33 @@ public function get_notification_footer( $post ) {
 	 * @param string $message_headers. (optional) Message headers
 	 */
 	public function send_email( $action, $post, $subject, $message, $message_headers = '' ) {
+		// Do nothing if the post isn't set.
+		$post_id = $post->ID;
```
alecgeatches: I'm not sure how this would happen, but it looks like you added it so there probably was an issue. I'd double check with an `isset()` just to avoid potential warnings:

```suggestion
		$post_id = isset( $post->ID ) ? $post->ID : false;
```

modules/notifications/notifications.php
```diff
@@ -217,19 +193,33 @@ public function get_notification_footer( $post ) {
...
-
-		if ( $recipients && ! is_array( $recipients ) ) {
-			$recipients = explode( ',', $recipients );
+		// Ensure the email address is set from settings.
+		if ( empty( VIP_Workflow::instance()->settings->module->options->email_address ) ) {
+			return;
 		}
 
-		// ToDo: Figure out best filters to add for the email.
+		$email_recipients[] = VIP_Workflow::instance()->settings->module->options->email_address;
```
alecgeatches: This sort of pattern is usually used when appending to an already-existing array. I'd suggest instantiating the variable here as an array to be more explicit:

```suggestion
		$email_recipients = [ VIP_Workflow::instance()->settings->module->options->email_address ];
```

modules/notifications/notifications.php
```diff
@@ -217,19 +193,33 @@ public function get_notification_footer( $post ) {
...
 
-		if ( ! empty( $recipients ) ) {
+		/**
+		 * Filter the email recipients
+		 *
+		 * @param array $email_recipients Array of email recipients
+		 * @param string $action Action being taken
```
alecgeatches: Can you add an example for what sort of thing might be in `$action` here? Like:

```suggestion
		 * @param string $action Action being taken, e.g. 'something' or 'something else'
```

modules/notifications/notifications.php
```diff
@@ -217,19 +193,33 @@ public function get_notification_footer( $post ) {
...
+		 * @param WP_Post $post Post object
+		 */
+		$email_recipients = apply_filters( 'vw_notification_email_recipients', $email_recipients, $action, $post );
+
+		// ToDo: Should we add a filter to allow for customizing the email subject, body and headers?
```
alecgeatches: I think a subject and body filter would be great here.

modules/shared/php/views/module-header.php
alecgeatches: FYI this file got blown away in https://github.com/Automattic/vip-workflow-plugin/pull/34, but we're [doing the same thing manually](https://github.com/Automattic/vip-workflow-plugin/blob/e99a1ef82de055e6207080fd0e587bbcec069aa1/modules/settings/views/settings.php#L20-L22) in the settings module.

modules/settings/settings.php
alecgeatches: For this one, you can wrap `$_POST['_wpnonce']` in `sanitize_key()` to get rid of this error. I found [this discussion about the wp_nonce sniff](https://github.com/WordPress/WordPress-Coding-Standards/issues/869#issuecomment-669119020) with a few well-known names.

modules/settings/settings.php
```diff
@@ -260,10 +222,12 @@ public function helper_settings_validate_and_save() {
...
 		if ( ! current_user_can( 'manage_options' ) || ! wp_verify_nonce( $_POST['_wpnonce'], VIP_Workflow::instance()->$module_name->module->options_group_name . '-options' ) ) {
 			wp_die( esc_html__( 'Cheatin&#8217; uh?' ) );
 		}
 
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized -- check is already done above
```
alecgeatches: This does look pretty dangerous at first glance, as we're getting an array from `$_POST` directly using a dynamic property name, and then and it's unclear that validation is actually being called due to the dynamic module setup:

https://github.com/Automattic/vip-workflow-plugin/blob/ceebcba6079ee00e4bbf1a37849f7b2c848d79c2/modules/settings/settings.php#L233-L236

For right now, can we remove the `if` wrapper on the code below this to make it obvious that we're calling `settings_validate()` on these values we get from `$_POST`?

alecgeatches (approved):
Left a few small notes, but otherwise this works great! Feel free to merge once you've had a look at the comments.

I also noticed some weird webhook behavior while testing, but because it's not related to the changes here I made a separate issue to track that in https://github.com/Automattic/vip-workflow-plugin/issues/36.

modules/settings/settings.php
ingeniumed: Good find - incredibly weird that it's needed but I understand why from that issue 👍🏾

modules/notifications/notifications.php
ingeniumed: This was actually something that was taken from Edit Flow, so I left it in there. I was actually leaning towards just removing it tbh considering how this code is triggered in the first place

modules/notifications/notifications.php
ingeniumed: I avoided it at first, so that we could be prescriptive about these filters. But I'll add it in as I'd been going back and forth on this.

modules/settings/settings.php
ingeniumed: Yip, we don't need that if check anymore so makes sense to kill it entirely

ingeniumed:
Fixed #36 in this PR as well. I moved the webhook logic to be under a cron command, and batched all the email recipients together for efficiency.

PR #38, opened 2024-09-18
Refactor the notifications module

ingeniumed:
## Description

I ended up splitting the notifications refactor, and writing up the tests PRs. There'll be another PR that follows this with the tests.

This refactors the notifications module to match what was done for the EM, and preview module. It converts to be isolated, static and adds types to the method definitions.

This also simplifies the webhook scheduling method, so that it can be used for cases beyond just post status updates. An example would be when a status update is rejected, it should be easy to call this method to send a rejection notice. 

I also noticed some previous refactoring items that were missed so I cleaned them up.

## Steps to Test

- Delete the options for notifications, and editorial metadata and ensure the install step is correctly triggered. 
- Ensure when post status changes happen the notifications are correctly sent.

### DIFF
--- a/class-workflow.php
+++ b/class-workflow.php
@@ -75,7 +75,7 @@ class VIP_Workflow {
 		// VIP Workflow base module
 		require_once VIP_WORKFLOW_ROOT . '/modules/shared/php/class-module.php';
 
-		$skip_module_dirs = [ 'shared', 'preview' ];
+		$skip_module_dirs = [ 'shared', 'preview', 'notifications', 'editorial-metadata' ];
 
 		// Scan the modules directory and include any modules that exist there
 		$module_dirs = scandir( VIP_WORKFLOW_ROOT . '/modules/' );
--- a/modules/notifications/notifications.php
+++ b/modules/notifications/notifications.php
@@ -1,48 +1,34 @@
 <?php
 /**
- * class Notifications
- * Email notifications for VIP Workflow and more
+ * The notifications module for sending email/webhook notifications.
+ *
+ * @package vip-workflow
  */
 namespace VIPWorkflow\Modules;
 
 use VIPWorkflow\VIP_Workflow;
-use VIPWorkflow\Modules\Shared\PHP\Module;
 use function VIPWorkflow\Modules\Shared\PHP\vw_draft_or_post_title;
 
-class Notifications extends Module {
-
-	public $module;
+class Notifications {
 
 	/**
-	 * Register the module with VIP Workflow but don't do anything else
+	 * Initialize the notifications module
 	 */
-	public function __construct() {
-
-		// Register the module with VIP Workflow
-		$this->module_url = $this->get_module_url( __FILE__ );
-		$args             = [
-			'module_url'            => $this->module_url,
-			'slug'                  => 'notifications',
-		];
-		$this->module     = VIP_Workflow::instance()->register_module( 'notifications', $args );
-	}
-
-	/**
-	 * Initialize the notifications class if the plugin is enabled
-	 */
-	public function init() {
+	public static function init(): void {
 		// Send notifications on post status change
-		add_action( 'transition_post_status', [ $this, 'notification_status_change' ], 10, 3 );
+		add_action( 'transition_post_status', [ __CLASS__, 'notification_status_change' ], 10, 3 );
+
 		// Schedule email sending
-		add_action( 'vw_send_scheduled_emails', [ $this, 'send_emails' ], 10, 4 );
+		add_action( 'vw_send_scheduled_emails', [ __CLASS__, 'send_emails' ], 10, 4 );
+
 		// Schedule webhook sending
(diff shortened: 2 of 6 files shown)


class-workflow.php
```diff
@@ -75,7 +75,7 @@ private function load_modules() {
 		// VIP Workflow base module
 		require_once VIP_WORKFLOW_ROOT . '/modules/shared/php/class-module.php';
 
-		$skip_module_dirs = [ 'shared', 'preview' ];
+		$skip_module_dirs = [ 'shared', 'preview', 'notifications', 'editorial-metadata' ];
```
alecgeatches, line 78: Good catch on `editorial-metadata`!

modules/notifications/notifications.php
```diff
@@ -1,48 +1,34 @@
...
 		// Schedule webhook sending
-		add_action( 'vw_send_scheduled_webhook', [ $this, 'send_to_webhook' ], 10, 4 );
+		add_action( 'vw_send_scheduled_webhook', [ __CLASS__, 'send_to_webhook' ], 10, 4 );
 	}
 
 	/**
 	 * Set up and send post status change notification email
 	 */
-	public function notification_status_change( $new_status, $old_status, $post ) {
+	public static function notification_status_change( string $new_status, string $old_status, \WP_Post $post ): void {
```
alecgeatches: `\WP_Post` works fine, but we can also `use WP_Post;` at the top of this file and the backslash would no longer be necessary.

alecgeatches: If you fix this, also see function signatures for `schedule_emails()` and `get_scheduled_datetime()`

modules/notifications/notifications.php
```diff
@@ -252,29 +244,19 @@ public function send_emails( $recipients, $subject, $message, $message_headers =
...
-	 * @param string $post_title Post title
-	 * @param string $old_status Old status of the post
-	 * @param string $new_status New status of the post
-	 * @param string $post_timestamp Timestamp of the post's last update
+	 * @param string $webhook_message Message to be sent to webhook
+	 * @param string $action Action being taken, eg. status-change
+	 * @param string $timestamp Timestamp of the message, eg. the time at which the post was updated
 	 */
-	public function schedule_webhook_notification( $current_user, $post_type, $post_id, $edit_link, $post_title, $old_status, $new_status, $post_timestamp ) {
+	public static function schedule_webhook_notification( string $webhook_message, string $action, string $timestamp ): void {
```
alecgeatches, line 254: Good improvement precomputing these values to make this monster function signature smaller.

vip-workflow.php
```diff
@@ -53,3 +53,4 @@
 // Modules
 require_once VIP_WORKFLOW_ROOT . '/modules/editorial-metadata/editorial-metadata.php';
 require_once VIP_WORKFLOW_ROOT . '/modules/preview/preview.php';
+require_once VIP_WORKFLOW_ROOT . '/modules/notifications/notifications.php';
```
alecgeatches, line 55: Can we make these alphabetical per section? No real purpose other than consistency.

alecgeatches (approved):
Left a couple of tiny comments, but the code and functionality works great. Love the refactor and improvements!

I have one related suggestion that may not be right for this PR, but I encountered during testing that I think would be useful. Our email for switching between statuses looks like this:

example-status-change-email

This is the email's subject:

```
[VIP Workflow] post Status Changed for "Fresh Post"
```

Beyond having some weird capitalization, I think this is missing the important information of what the new posts' status is. I image if a customer uses this notification feature, they'll probably filter out some of the messages and primarily be interested when a post hits a couple of interesting statuses. I think we could improve this by adding the new status to the subject:


```
[VIP Workflow] "Fresh Post" Post moved to "Status 2"
```

With this setup, if I'm only interested in "Status 2" or I want to filter that into a particular folder, it's a lot easier. It looks like we only need to change this line:

https://github.com/Automattic/vip-workflow-plugin/blob/57fa6d3a6fdfa428dae38003945220caeef9ab85/modules/notifications/notifications.php#L121-L122

We can also [`ucfirst`](https://www.php.net/manual/en/function.ucfirst.php) the actual name related to the `$post_type` to avoid weird capitalization. The other generated `$subject`s are unique and easier to filter for, so I only think we'd want to change this one.

modules/notifications/notifications.php
ingeniumed: Good catch, fixed all this up

modules/notifications/notifications.php
ingeniumed: Yip, makes it easier to use for the approve/reject code too

vip-workflow.php
ingeniumed: Done

ingeniumed:
Good point on the email subject, I've improved that now

ingeniumed:
Screenshot 2024-09-19 at 10 33 19 am

PR #42, opened 2024-09-25
Add restricted status UI for users

alecgeatches:
## Description

Adds workflow settings UI to restrict statuses to specified privileged users:

restricted-by-user-ui

There are a couple of important things missing from this PR that will be added in future PRs:

- The ability to select privileged roles, like "Administrator" along with individual users. This will be separated as this PR already has a lot of changes.
- Disallowing status movements dependent on user and role. Currently, this is configurable, but does not have any effect on the editor button.

## Steps to Test

1. Check out PR (`add/restricted-status-ui`).
2. Go to Admin -> VIP Workflow -> VIP Workflow.
3. Click the edit pencil icon next to a status.
4. Next to the "This status is restricted" label, click the toggle button.
5. Under "ALLOWED USERS", enter the display names or usernames of some users. You may need to create some for testing under Admin -> Users -> Add New Users.
6. Click save and reload the page.
7. Edit the changed status again. Ensure the status now has the "This status is restricted" toggle enabled by default with the selected set of users.

## Other notes

This feature ended up being pretty complicated, mostly because we want to display and store user data in several ways, depending on the context:

user-suggestion-parts

1. The first red box is how we display already-selected users, by their `user_login`.
2. The second red box is how we display user suggestions, with both their display name and `user_login`.
3. On save we store the `user_id` associated with the `user_login` selected by the user.

It's up to the code to suggest, store, and save with different values, which adds some complexity. I'll add some notes below pointing out where this happens in code.


modules/custom-status/custom-status.php
```diff
@@ -87,10 +91,9 @@ public function init() {
 		// Pagination for custom post statuses when previewing posts
 		add_filter( 'wp_link_pages_link', [ $this, 'modify_preview_link_pagination_url' ], 10, 2 );
 
-		// REST endpoints
-		CustomStatusEndpoint::init();
-
 		add_filter( 'user_has_cap', [ $this, 'remove_or_add_publish_capability_for_user' ], 10, 3 );
+
+		add_filter( 'vw_custom_status_ui_additions', [ $this, 'add_custom_status_ui_properties' ] );
```
alecgeatches: Statuses are stored with `required_user_ids`, but this isn't enough to display corresponding `user_login` string in the UI. This filter is used for custom status preprocessing on initial page load as well as REST responses to fill in `required_user_logins` to supply a value that can be displayed in the UI.

modules/custom-status/custom-status.php
```diff
@@ -134,10 +137,9 @@ public function install() {
 			[
 				'term' => __( 'Pending Review' ),
 				'args' => [
-					'slug'               => 'pending',
-					'description'        => __( 'Post needs to be reviewed by an editor.', 'vip-workflow' ),
-					'position'           => 5,
-					'is_review_required' => true,
```
alecgeatches, line 140: `is_review_required` has been removed, as this can be computed from the presence of `required_user_ids`, and later other conditions like roles.

modules/custom-status/custom-status.php
```diff
@@ -490,13 +510,19 @@ public function add_custom_status( $term, $args = [] ) {
...
-		// Populate the inserted term with the new values, or else only the term_taxonomy_id and term_id are returned.
 		if ( is_wp_error( $inserted_term ) ) {
 			return $inserted_term;
-		} else {
-			$inserted_term = $this->get_custom_status_by( 'id', $inserted_term['term_id'] );
 		}
 
+		// Add term meta
+		if ( [] !== $required_user_ids ) {
+			update_term_meta( $inserted_term['term_id'], 'required_user_ids', $required_user_ids );
```
alecgeatches: `required_user_ids` passed in through `$args` will be stored in term meta here.

modules/custom-status/custom-status.php
```diff
@@ -662,6 +690,11 @@ public function get_custom_statuses() {
 					$status->$key = $value;
 				}
 			}
+
+			// Add term meta
+			$required_user_ids         = get_term_meta( $status->term_id, 'required_user_ids', true );
+			$status->required_user_ids = '' === $required_user_ids ? [] : $required_user_ids;
```
alecgeatches: The term's `required_user_ids` property is filled in automatically here from term meta.

modules/custom-status/lib/components/modals/create-edit-custom-status-modal.js
```diff
@@ -34,7 +46,7 @@ export default function CreateEditCustomStatusModal( { customStatus, onCancel, o
 		const data = {
 			name,
 			description,
-			is_review_required: isReviewRequired,
+			required_user_logins: requiredUserLogins,
```
alecgeatches: Note that the user restrictions component only cares about user logins, and sends these to the REST endpoint. The REST endpoint then converts these to user IDs for storage. User logins are there for users to understand, and user IDs are better for consistency when stored.

modules/custom-status/rest/custom-status-endpoint.php
```diff
@@ -349,3 +387,5 @@ public static function get_reorder_url() {
 		return rest_url( sprintf( '%s/%s', VIP_WORKFLOW_REST_NAMESPACE, 'custom-status/reorder' ) );
 	}
 }
+
+CustomStatusEndpoint::init();
```
alecgeatches, line 374: I updated this endpoint and all others in our plugin to `::init()` themselves. There was really no reason to `::init()` manually in the including class.

modules/custom-status/custom-status.php
```diff
@@ -1314,5 +1347,17 @@ public function fix_post_row_actions( $actions, $post ) {
 		$actions['view'] = '<a href="' . esc_url( $preview_link ) . '" title="' . esc_attr( sprintf( __( 'Preview &#8220;%s&#8221;' ), $post->post_title ) ) . '" rel="permalink">' . __( 'Preview' ) . '</a>';
 		return $actions;
 	}
+
+	public function add_custom_status_ui_properties( $custom_status ) {
+		// Add 'required_user_logins' data with display names for status restrictions UI
+		$required_user_ids = $custom_status->required_user_ids ?? [];
+
+		$custom_status->required_user_logins = array_filter(array_map( function ( $user_id ) {
```
ingeniumed: If you port over the `get_term` and `get_terms` filters from #41 you can move this code there. It would be targeted at just the custom status taxonomy, and would ensure that the data was available. The only change would be that the data would be under the meta field, rather than being directly accessible.

Happy to also do that in a follow up PR from my end tbh.

modules/custom-status/custom-status.php
ingeniumed: Noticed that you used `update_term_meta` here instead of `add_term_update`. Was that on purpose? Figured that, that would be reserved for the `update_custom_status` method instead?

modules/custom-status/custom-status.php
```diff
@@ -1314,5 +1347,17 @@ public function fix_post_row_actions( $actions, $post ) {
...
+			$user = get_user_by( 'ID', $user_id );
+			return $user instanceof WP_User ? $user->user_login : false;
```
ingeniumed: I do like the safety check here. I'd wanna do more, just to be safe:

- Hook into `do_action( 'deleted_user', $id, $reassign, $user );` to ensure that the user IDs in the term_meta is always correct.
- In the event that a user is not available or is not the right type, that ID should immediately be dropped.

modules/custom-status/rest/user-lookup-endpoint.php
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+namespace VIPWorkflow\Modules\CustomStatus\REST;
+
+use WP_REST_Request;
+use WP_User_Query;
+
+defined( 'ABSPATH' ) || exit;
+
+class UserLookupEndpoint {
```
ingeniumed: We can replace this whole thing with a local call within JS itself using `wp`.

```javascript
wp.apiFetch( { path: '/wp/v2/users?search=test' } )
```

That'll give back the list of matches, and each entry has the entire user object. This way it would save a roundtrip and would be nice and speedy.

modules/custom-status/rest/custom-status-endpoint.php
```diff
@@ -43,16 +44,22 @@ public static function register_routes() {
...
-				'description'        => [
+				'description'          => [
 					'default'           => '',
 					'sanitize_callback' => function ( $param ) {
 						return stripslashes( wp_filter_nohtml_kses( trim( $param ) ) );
 					},
 				],
-				'is_review_required' => [
-					'default'           => false,
+				'required_user_logins' => [
```
ingeniumed: If we go with the idea of calling the user endpoint via wp api, I think we should switch this to be user IDs instead. Then, we can simply validate they exist with a quick function call or query and move on to using them. We would save another call to convert them to IDs as well. It'll be a lot quicker to do a quick lookup on the IDs.

I can see right now a delay in the whole process, compared to before and that's only with 2 users on a status out of a total user set of 4 users.

modules/custom-status/rest/custom-status-endpoint.php
```diff
@@ -260,17 +276,24 @@ public static function handle_update_status( WP_REST_Request $request ) {
...
 		// ToDo: Ensure that we don't do an update when the name and description are the same as the current status
-		$update_status_result = $custom_status_module->update_custom_status( $term_id, $args );
+		$updated_status = $custom_status_module->update_custom_status( $term_id, $args );
+
+		/**
+		* Add additional data to status terms used in the UI
+		*
+		* @param WP_Term $updated_status An updated status term.
+		*/
+		$updated_status = apply_filters( 'vw_custom_status_ui_additions', $updated_status );
```
ingeniumed: I think it'd be better to move this into the `update_custom_status` so that you get the entire object after a save or update.

ingeniumed (changes requested):
Nice work on adding this in! There's a couple of changes I think we should make still:

- Switch the user search call to be done via JS so it's faster and cleaner
- Speed up the save process by reducing the number of validations made to the DB. Sending just the User IDs down means we only need to validate it once and then it's saved.
- Process user deletes by ensuring that meta value is kept updated always. Added to that, drop any user that doesn't match the right type or may not exist.

modules/custom-status/rest/custom-status-endpoint.php
ingeniumed: Note: I do like the idea of adding the username instead of just the IDs as it saves a call. Adding it to another transient meta might make sense so it's known that it's not actually stored in the DB. Overriding the `get_term` and `get_terms` call should do the trick.

modules/custom-status/custom-status.php
ingeniumed: Scratch that, don't do this approach. I assumed that I'd be able to set different priorities on editorial metadata and custom status so that I could have different get_term and get_terms calls. But that has failed and there is a way to work around this, which is to have the two modules share code. That defeats the purpose of trying to keep them isolated. So its better to just have the get and get_by calls within the individual modules be the source of truth for any retrievals.

modules/custom-status/custom-status.php
alecgeatches: In general the `update_*` functions in WordPress (`update_post_meta()`, `update_user_meta()`, etc) can add  meta, but they will also update an existing value if it exists. [`add_term_meta()`](https://developer.wordpress.org/reference/functions/add_term_meta/) can do the same if the `$single` parameter is set to `true`, but by default it will add a second value for a meta key if it already exists.
  
Using an `update_*` function is usually the correct choice unless you'd like non-`$single` data, like several values saved with the same key. `update_post_meta()` assumes 1:1 data, and if I saw a `add_post_meta()` I'd look closely for a 1:many relationship.

modules/custom-status/rest/user-lookup-endpoint.php
alecgeatches: Good thinking! I completely forgot about the native endpoint.

modules/custom-status/rest/custom-status-endpoint.php
alecgeatches: I also think we should switch to IDs here just for a better design, I'm not super happy with using logins in the REST API for UI reasons. That being said, I don't think verifying users is going to have a large overhead. I tested execution time for `convert_user_logins_to_user_ids()` in the `required_user_logins` parameter validation step like this:

```php
'required_user_logins' => [
    'default'           => [],
    'validate_callback' => function ( $param ) {
        $user_logins = array_map( 'sanitize_text_field', $param );

        $start_time     = microtime( true );
        $user_ids       = self::convert_user_logins_to_user_ids( $user_logins );
        $end_time       = microtime( true );
        $execution_time = $end_time - $start_time;
        error_log( 'Execution time: ' . $execution_time );

        return count( $user_ids ) === count( $user_logins );
    },
    'sanitize_callback' => function ( $param ) {
        return array_map( 'sanitize_text_field', $param );
    },
],
```

The same code was also added to the second execution of `convert_user_logins_to_user_ids()` in `handle_create_status`. This was tested locally, with 10 users selected out of a total ~1,100 test users. The results:

```
NOTICE: PHP message: Execution time: 0.0030879974365234
NOTICE: PHP message: Execution time: 0.0006120204925537109
```

The first invocation took about 3ms, and the second about 0.6ms. In a real environment DB times will certainly be worse, but with caching the overhead is still very small. We'll likely incur a similar overhead verifying IDs as we do logins.

---

All that said, I think IDs make more sense anyway, so I'll have a go at converting over to that.

alecgeatches:
@ingeniumed Thank you for the code review, you had a lot of good suggestions! This is ready for review again. I've made these changes:

1. Dropped `UserLookupEndpoint` in favor of the native API endpoint. Glad you caught that!

3. Changed the REST endpoint (and related UI code) to send and receive user IDs instead of logins. I don't believe this has a big speed advantage, but it feels cleaner and more proper. The reason I used logins originally was due to UI display which wasn't a good design anyway.
4. Instead of using a filter, modified `get_custom_statuses()` to fill in user metadata from `required_user_ids`. This computed property is called `required_users`, which only includes an ID and username, but has been renamed from `required_user_logins` so it's not necessarily tied to usernames if we wish to display something else in the UI later.
5. Moved meta manipulation for user IDs into a separate [`RequiredUserIds`](https://github.com/Automattic/vip-workflow-plugin/blob/0e400a67e81e890e627e63e0ffde21b81ebec751/modules/custom-status/meta/required-user-ids.php) class. I think this is a pattern we can probably reuse in #43. The main benefit is that the metadata is encapsulated in the class, including the term meta key and getter/setter functions. Even better, meta-specific filters (like `delete_user`) can also be nicely encapsulated in the meta class itself. A nice alternative to adding more `const`s and callback code in the `Custom_Status` module!
6. Ensured user IDs are kept up to date on user delete (also in [`RequiredUserIds`](https://github.com/Automattic/vip-workflow-plugin/blob/0e400a67e81e890e627e63e0ffde21b81ebec751/modules/custom-status/meta/required-user-ids.php)), including reassignment in the class above. Great thinking here too! To test the reassignment capabilities, run `wp_delete_user( <user_id_to_delete>, <user_id_to_assign_to> );`, refresh the custom status UI, and see that the user has updated correctly.
7. Discovered a UI/UX bug where the "This status is restricted" toggle didn't work properly when users were set, as it was computed from set restriction values. The only way to remove the restriction was to manually remove each user. Here's the buggy behavior:

    unrestrict-status-bug

    ---

    This was updated to respect the toggle. If a user turns off restrictions and then saves the custom status, the restrictions are removed:

    remove-status-restrictions

4. Discovered another bug that happened when typing `,` or the enter key when no user suggestions were present. This caused the whole React app to crash. This has been fixed in https://github.com/Automattic/vip-workflow-plugin/pull/42/commits/50930043ada2390a0a4b8535ec72d2fb0cf24f34 to ignore entries when no matching user suggestion is present.

modules/custom-status/meta/required-user-ids.php
```diff
@@ -0,0 +1,61 @@
...
+				continue;
+			}
+
+			if ( null === $reassigned_user_id ) {
+				// Remove the deleted user
+				unset( $required_user_ids[ $deleted_user_index ] );
+				$required_user_ids = array_values( $required_user_ids );
+			} else {
+				// Replace the deleted user with the reassigned user
+				$required_user_ids[ $deleted_user_index ] = $reassigned_user_id;
```
ingeniumed, line 36: Beauty! This would be super useful for customers as they don't need to do this.

modules/custom-status/rest/custom-status-endpoint.php
```diff
@@ -329,6 +337,19 @@ public static function handle_reorder_status( WP_REST_Request $request ) {
 		return rest_ensure_response( $custom_status_module->get_custom_statuses() );
 	}
 
+	// Utility functions
+
+	private static function is_valid_user_ids( $user_ids ) {
+		foreach ( $user_ids as $user_id ) {
+			$user = get_user_by( 'id', $user_id );
```
ingeniumed, line 344: If we use [this](https://stackoverflow.com/a/62574404), this can be done in one call so it'd be a bit more efficient and faster.

ingeniumed (approved):
Minor performance improvement, otherwise this looks good to me.

I can make that change in my PR tbh.

modules/custom-status/rest/custom-status-endpoint.php
alecgeatches: Good thinking! I can update that in another PR.

PR #55, opened 2024-10-02
Update the design for the required fields on a custom status

ingeniumed:
## Description

This PR is meant to update the design for configuring the required fields on a status, using the mockups proposed by @jarekmorawski. Check out #44 for an example of this.

The changes I've made:

- Implement a radio control for turning the user restriction on and off.
- Change the labels to be clearer and drop the help labels when they aren't necessary.

Note: The rules engine is not in place for editorial metadata just yet. Once it is, that section would change a bit more so keep that in mind when reviewing this.

Screenshot 2024-10-02 at 12 07 35 pm
Screenshot 2024-10-02 at 12 07 45 pm
Screenshot 2024-10-02 at 12 07 52 pm
Screenshot 2024-10-02 at 12 07 58 pm
Screenshot 2024-10-02 at 12 08 37 pm

## Steps to Test

- Update and Create a status with editorial metadata fields, as well as users to see the difference.


modules/custom-status/lib/components/modals/create-edit-custom-status-modal.js
```diff
@@ -55,8 +53,8 @@ export default function CreateEditCustomStatusModal( {
 	// Modal properties
 	const [ error, setError ] = useState( null );
 	const [ isRequesting, setIsRequesting ] = useState( false );
-	const [ isRestrictedSectionVisible, setIsRestrictedSectionVisible ] = useState(
-		requiredUsers.length > 0 || requiredMetadatas.length > 0
+	const [ areRestrictedUsersSet, setAreRestrictedUsersSet ] = useState(
```
ingeniumed, line 56: Not the biggest fan of setting text here to control the radio control, so happy to change this. This seemed like the best way imo for now.

ingeniumed:
Note: I've dropped roles from the verbiage but the screenshots still reflect the old verbiage. Just noting that, as we will add roles support in the future.

jarekmorawski:
It's looking good! The only part I'm unsure of is the required metafields label. It isn't clear to me what the field does. Would it make sense to change it to something like below?

image

Note that it's the inversion of the logic suggested in the new design where users would set visibility requirements on the succeeding status (e.g., 'Legal review' would only be visible if the 'Needs legal review' metafield is filled), instead of the current one (e.g., only move forward if X). I can't say if any of the options is better in the long term, but it does impact the UX and may cause potential complications later on. cc @smithjw1

ingeniumed:
Something I had gone back and forth on, was keeping that help text as it makes it clear. I ended up removing that help text in favour of the label being descriptive enough. I'll add it back in, and shift the required metadata fields above the users. Given the user input is hidden, it'll prevent the metadata fields from being pushed down.

We would want to keep the users and metadata fields requirement being applied on the current status rather than the next one, as that keeps the whole thing simple.

ingeniumed:
Screenshot 2024-10-02 at 12 08 37 pm

I've made the above change

ingeniumed:
Screenshot 2024-10-02 at 8 03 50 pm

Issue #49, opened 2024-10-01
Move the core hacks to their own class


ingeniumed:
Once #45 is done, move the core hacks to their own class from custom status. That way, we know exactly what improvements we want in core and can reduce the old code we have in our plugin.

Issue #46, opened 2024-10-01
Raise min WP and PHP versions


ingeniumed:
Set the minimum PHP version to 8.1 or 8.2 and the minimum WP version to 6.4.

PR #59, opened 2024-10-04
Refactor the remaining modules

### LINKED ISSUE #45: Refactor custom status
This is meant to refactor custom status module the same way as #34. As part of this, ensure that the module management system taken from Edit Flow is deleted entirely and the remaining non static methods are refactored or deleted entirely. 

The end goal should be to remove the remaining pieces keeping the current module management system in place.

Bonus: See if the folder structure could be changed as well.

ingeniumed:
## Description

This is a big PR and needs to be reviewed and tested thoroughly. I have refactored the remaining modules to be independent, so all the remaining Edit Flow logic is gone. I've left the folder structure in place as that felt like it'd be way too much. That would be the next item on the list imo.

I have also updated the minimum PHP version to 8.1 and the minimum WordPress version to 6.4.

Module specific changes:

Custom Status: I've made the pending review status mandatory. That felt like it's necessary because some weird bugs were beginning to show up where the pending status was being duplicated. If no statuses are present, we assume just draft and pending are allowed and those are the default statuses in core to begin with. So I've left that as is. I've also shifted all the core hacks away to their own class.

Settings: Everything is isolated to this module, but all options related operations happen through one utility. It ensures that the default option values exist and if we do enable migrations that can go there. I've also updated the settings updated notice to be an admin banner similar to our regular notice.

Items remaining that we need to tackle for the future:

Supporting post types beyond the two default ones is not present right now. Technically we didn't have it working properly before, because we deleted the Edit Flow code. So we need to that back in to ensure we support that.

Fixes #45, #46, and #49 

## Steps to Test

- Setup the plugin from scratch (delete everything)
- Ensure all the defaults get setup
- Ensure everything works


modules/custom-status/custom-status.php
```diff
@@ -178,20 +124,16 @@ public static function register_custom_status_taxonomy(): void {
 	 * Makes the call to register_post_status to register the user's custom statuses.
 	 * Also unregisters pending, in case the user doesn't want them.
 	 */
-	public function register_custom_statuses() {
+	public static function register_custom_statuses(): void {
 		global $wp_post_statuses;
 
-			// Users can delete the pending status if they want, so let's get rid of that
-			// It'll get re-added if the user hasn't "deleted" them
-			unset( $wp_post_statuses['pending'] );
```
ingeniumed, line 187: I ended up removing this, and making the pending status as required as well. We assume right now that both drafts and pending review are going to be required, and won't be removed. If all statuses are removed, those two are given back. I also ran into a minor edge case where pending review was being created twice with different slugs. So by doing this, we are mirroring the behaviour that happens when no statuses exist. Based on what I saw in edit flow, those restrictions were put on purpose so this is better to do imo.

modules/custom-status/custom-status.php
```diff
@@ -678,23 +726,26 @@ public function delete_custom_status( int $status_id ): bool|WP_Error {
 
 		$old_status_slug = $old_status->slug;
 
-		if ( $this->is_restricted_status( $old_status_slug ) || 'draft' === $old_status_slug ) {
-			return new WP_Error( 'restricted', __( 'Restricted status ', 'vip-workflow' ) . '(' . $old_status->name . ')' );
+		// Don't allow changing the name or slug of the Draft or pending status as they are core statuses
+		$banned_statuses = [ 'draft', 'pending' ];
+
+		if ( self::is_restricted_status( $old_status_slug ) || in_array( $old_status->slug, $banned_statuses ) ) {
+			return new WP_Error( 'restricted', __( 'Restricted status ', 'vip-workflow' ) . '(' . $old_status->name . ') cannot be deleted.' );
```
ingeniumed: Changed this message to be a lot more clearer

modules/custom-status/custom-status.php
```diff
@@ -582,25 +627,28 @@ public function add_custom_status( array $args ): WP_Term|WP_Error {
...
 		// Prevent user from changing draft name or slug
-		if ( 'draft' === $old_status->slug
+		if ( in_array( $old_status->slug, $banned_statuses )
 		&& (
 			( isset( $args['name'] ) && $args['name'] !== $old_status->name )
 			||
 			( isset( $args['slug'] ) && $args['slug'] !== $old_status->slug )
 		) ) {
-			return new WP_Error( 'invalid', __( 'Changing the name and slug of "Draft" is not allowed', 'vip-workflow' ) );
+			return new WP_Error( 'restricted', __( 'Changing the name and slug of a restricted status ', 'vip-workflow' ) . '(' . $old_status->name . ') is not allowed.' );
```
ingeniumed: Changed this message to be a lot more clearer

modules/settings/settings.php
```diff
@@ -8,118 +8,78 @@
...
 	/**
-	 * Add settings JS to the settings page
+	 * Enqueue resources that we need in the admin settings page
+	 *
+	 * @access private
 	 */
-	public function action_admin_enqueue_scripts() {
-		if ( VIP_Workflow::is_settings_view_loaded( self::SETTINGS_SLUG ) ) {
-			wp_enqueue_script( 'vip-workflow-settings-js', $this->module_url . 'lib/settings.js', array( 'jquery' ), VIP_WORKFLOW_VERSION, true );
+	public static function action_admin_enqueue_scripts(): void {
```
ingeniumed, line 54: Made the settings script loading similar to how we usually do it.

modules/settings/settings.php
```diff
@@ -158,28 +118,78 @@ public function helper_option_custom_post_type( $args = array() ) {
...
-			. esc_attr( $this->module->options_group_name ) . '[post_types][' . esc_attr( $post_type ) . ']"';
-			if ( isset( $this->module->options->post_types[ $post_type ] ) ) {
-				checked( $this->module->options->post_types[ $post_type ], 'on' );
+			. esc_attr( OptionsUtilities::get_module_options_key( self::SETTINGS_SLUG ) ) . '[post_types][' . esc_attr( $post_type ) . ']"';
+			$the_post_types = OptionsUtilities::get_module_option_by_key( self::SETTINGS_SLUG, 'post_types' );
+			if ( isset( $the_post_types[ $post_type ] ) ) {
+				checked( $the_post_types[ $post_type ], 'on' );
 			}
-			// Defining post_type_supports in the functions.php file or similar should disable the checkbox
-			disabled( post_type_supports( $post_type, $this->module->post_type_support ), true );
```
ingeniumed, line 166: Noting that this needs to be added back in. Right now it was half baked in trunk, so it didn't make sense leaving it in imo.

modules/settings/views/settings.php
```diff
@@ -18,15 +21,20 @@
 		<h3>
 			<?php esc_html_e( 'Configure VIP Workflow settings.', 'vip-workflow' ); ?>
 			<?php if ( $message_slug && isset( $messages[ $message_slug ] ) ) { ?>
-				<?php printf( '<span class="vip-workflow-updated-message vip-workflow-message">%s</span>', esc_html( $messages[ $message_slug ] ) ); ?>
+				<?php
```
ingeniumed, line 24: This is a method specific to 6.4 so I've updated our minimum versions to 6.4 along with our php version to 8.1

modules/shared/php/options-utilities.php
```diff
@@ -8,18 +8,45 @@
...
 
 use stdClass;
-use VIPWorkflow\VIP_Workflow;
 
 class OptionsUtilities {
+	const OPTIONS_GROUP = 'vip_workflow_';
+	const OPTIONS_GROUP_NAME = 'vip_workflow_options';
+
+	// Its key to centralize these here, so we can easily update them in the future
+	private const DEFAULT_OPTIONS = [
```
ingeniumed: I've kept the default options here but I'd like to move this eventually to a typed class that we can rely on easily.

modules/shared/php/options-utilities.php
```diff
@@ -8,18 +8,45 @@
...
 	 * @return object The set of saved module options for this module, or an empty stdClass if none are found
 	 */
-	public static function get_module_options( string $module_slug ): object {
+	public static function get_module_options( string $module_slug ): object|null {
 		$module_options_key = self::get_module_options_key( $module_slug );
-		return get_option( $module_options_key, new stdClass() );
+		$module_options = get_option( $module_options_key, new stdClass() );
+
+		// Ensure all default options are set
+		foreach ( self::DEFAULT_OPTIONS as $key => $value ) {
```
ingeniumed: This is the new location for the default options being saved.

tests/modules/custom-status/meta/test-required-metadata-id-handler.php
```diff
@@ -22,12 +23,20 @@ protected function setUp(): void {
...
-		VIP_Workflow::instance()->custom_status->install();
+		CustomStatus::setup_install();
+	}
+
+	// tear down after class
+	protected function tearDown(): void {
+		parent::tearDown();
+
+		// Reset all module options
+		OptionsUtilities::reset_all_module_options();
```
ingeniumed: Due to our new static refactoring, this is necessary to do. It ensures each test doesn't conflict with the other. Otherwise, one test could trigger a module's install step and the other one wouldn't be able to do the same as it's already done.

modules/shared/php/core-hacks.php
```diff
@@ -0,0 +1,398 @@
...
+namespace VIPWorkflow\Modules\Shared\PHP;
+
+use VIPWorkflow\Modules\CustomStatus;
+use VIPWorkflow\Modules\Shared\PHP\HelperUtilities;
+
+use function VIPWorkflow\Modules\Shared\PHP\_vw_wp_link_page;
+
+use WP_Post;
+
+class CoreHacks {
```
ingeniumed, line 14: This is the central class with all the core hacks

modules/custom-status/custom-status.php
```diff
@@ -41,116 +37,47 @@ class Custom_Status extends Module {
...
-		$this->module_url = $this->get_module_url( __FILE__ );
-		// Register the module with VIP Workflow
-		$args         = [
-			'module_url'        => $this->module_url,
-			'slug'              => 'custom-status',
-			'configure_page_cb' => 'print_configure_view',
-		];
-		$this->module = VIP_Workflow::instance()->register_module( 'custom_status', $args );
-	}
+	const PUBLISHED_STATUSES = array(
```
alecgeatches: Minor: Looks like `array()`s have almost entirely been removed from code, regexing `\Warray\(` shows 16 results left. Would you mind removing these as well?

alecgeatches: Additionally, it looks like `PUBLISHED_STATUSES` is only used in one place in `core-hacks.php` so we might as well move to that file as well.

modules/custom-status/custom-status.php
alecgeatches: I'd prefer we allow the "Pending Review" status to be removed. At present you're not even allowed to rename it which feels restrictive for some of the workflows we may expect that don't want to use that naming convention. I think it would be even better if we could also remove or rename "Draft" for flexibility, but if we require any statuses I think "Draft" is the only one broad enough to be it.

modules/custom-status/custom-status.php
```diff
@@ -232,36 +153,126 @@ public function disable_custom_statuses_for_post_type( $post_type = null ) {
 		return false;
 	}
 
-	public function add_admin_menu() {
+	/**
+	 * Checks for the current post type
+	 *
+	 * @return string|null $post_type The post type we've found, or null if no post type
+	 */
+	public static function get_current_post_type(): ?string {
```
alecgeatches: We could probably move this to `HelperUtilities`. Definitely similar to the existing `get_supported_post_types()` in that module. It's currently only used in this file (and could also be made `private` instead), but I think it's a useful enough function to live somewhere shared.

modules/shared/php/options-utilities.php
```diff
@@ -8,18 +8,45 @@
...
+			if ( ! isset( $module_options->$key ) ) {
+				$module_options->$key = $value;
+			}
+		}
+
+		return $module_options;
+	}
+
+	public static function get_module_option_by_key( string $module_slug, string $key ): string|array|bool|null {
```
alecgeatches: It seems like this is always called with `Settings::SETTINGS_SLUG`. I don't believe we're expecting to store options in separate modules, so I think it would make sense to simplify and move that constant here, and also remove the `$module_slug` parameter. Callers could then just ask for the option:

```php
OptionsUtilities::get_option( 'publish_guard' )
```

modules/custom-status/custom-status.php
```diff
@@ -582,25 +606,28 @@ public function add_custom_status( array $args ): WP_Term|WP_Error {
...
 		} else if ( ! $old_status ) {
 			return new WP_Error( 'invalid', __( "Custom status doesn't exist.", 'vip-workflow' ) );
 		}
 
 		// Reset our internal object cache
-		$this->custom_statuses_cache = [];
+		self::$custom_statuses_cache = [];
+
+		// Don't allow changing the name or slug of the Draft or pending status as they are core statuses
+		$banned_statuses = [ 'draft', 'pending' ];
```
alecgeatches: As with [the prior comment](https://github.com/Automattic/vip-workflow-plugin/pull/59#discussion_r1792538988), I'd love it if we could at least make "pending" deleteable.

modules/custom-status/custom-status.php
alecgeatches: I like this improvement! We can utilize i18n functions here to avoid string concats and i18n mixing:

```suggestion
			// translators: %s: Post type, like "Draft"
			return new WP_Error( 'restricted', sprintf( __( 'Changing the name and slug of a restricted status (%s) is not allowed.', 'vip-workflow' ), $old_status->name ) );
```

alecgeatches: Also the same thing below in `delete_custom_status()`!

modules/custom-status/custom-status.php
```diff
@@ -678,23 +705,26 @@ public function delete_custom_status( int $status_id ): bool|WP_Error {
 
 		$old_status_slug = $old_status->slug;
 
-		if ( $this->is_restricted_status( $old_status_slug ) || 'draft' === $old_status_slug ) {
-			return new WP_Error( 'restricted', __( 'Restricted status ', 'vip-workflow' ) . '(' . $old_status->name . ')' );
+		// Don't allow changing the name or slug of the Draft or pending status as they are core statuses
+		$banned_statuses = [ 'draft', 'pending' ];
+
+		if ( self::is_restricted_status( $old_status_slug ) || in_array( $old_status->slug, $banned_statuses ) ) {
```
alecgeatches: I wonder if we can just move `draft` (and maybe `pending`) to `is_restricted_status()`? If we always ban it, it seems clearer to do in that function.

modules/custom-status/custom-status.php
```diff
@@ -763,19 +793,42 @@ public function get_custom_statuses(): array {
...
 
 		return $statuses;
 	}
 
+	/**
+	 * Get core's 'draft' and 'pending' post statuses, but include our special attributes
+	 *
+	 * @return array
+	 */
+	private static function get_core_post_statuses(): array {
```
alecgeatches: Similar point as some other comments, but I'm not sure why we need "core" post statuses, any more than we need core EM fields. If we have 0 or 1 statuses, it would be great if we could handle that gracefully in code. Don't need to do that in this PR, just a general thought.

modules/custom-status/meta/position-handler.php
```diff
@@ -29,9 +29,9 @@ public static function init(): void {
 	 * @return array The updated meta keys
 	 */
 	public static function add_position( array $term_meta, WP_Term $custom_status ): array {
-		$position = MetaCleanupUtilities::get_int( $custom_status->term_id, Custom_Status::METADATA_POSITION_KEY );
+		$position = MetaCleanupUtilities::get_int( $custom_status->term_id, CustomStatus::METADATA_POSITION_KEY );
```
alecgeatches, line 32: I like that you've split these filter side-effects out into a new file. My ideal next step would be to move the `METADATA_POSITION_KEY` constant into this file, and also add a `Position::get()` and `Position::set()` wrappers so `CustomStatus` doesn't own or need to know the details about how we store values on a term, and can just call into these meta files.

modules/settings/settings.php
```diff
@@ -7,119 +7,79 @@
...
-			<?php
-	}
-
 	/**
 	 * Register the settings for the module
 	 */
-	public function register_settings() {
-		add_settings_section( $this->module->options_group_name . '_general', false, '__return_false', $this->module->options_group_name );
+	public static function register_settings(): void {
+		add_settings_section( OptionsUtilities::get_module_options_general_key( self::SETTINGS_SLUG ), false, '__return_false', OptionsUtilities::get_module_options_key( self::SETTINGS_SLUG ) );
```
alecgeatches: Can we precompute these in variables and shorten these lines a bit?

```suggestion
		$settings_section_id = OptionsUtilities::get_module_options_general_key( self::SETTINGS_SLUG );
		$settings_section_page = OptionsUtilities::get_module_options_key( self::SETTINGS_SLUG );
		
		add_settings_section( $settings_section_id, false, '__return_false', $settings_section_page );
```

We may also just remove `self::SETTINGS_SLUG` entirely if it's always the only parameter [like this other comment](https://github.com/Automattic/vip-workflow-plugin/pull/59#discussion_r1792544850).

modules/settings/settings.php
alecgeatches: Can we make a separate issue for this? I'm not clear on what's baked/half-baked/unimplemented as well.

modules/settings/settings.php
```diff
@@ -208,33 +201,32 @@ public function settings_validate( $new_options ) {
...
 	 * This method is called automatically/ doesn't need to be registered anywhere
 	 */
-	public function helper_settings_validate_and_save() {
-
+	public static function helper_settings_validate_and_save(): bool {
 		if ( ! isset( $_POST['action'], $_POST['_wpnonce'], $_POST['option_page'], $_POST['_wp_http_referer'], $_POST['submit'] ) || ! is_admin() ) {
 			return false;
 		}
 
 		if ( 'update' != $_POST['action']
```
alecgeatches: We can add some `===`/`!==` in this function for all of the string comparisons.

modules/custom-status/custom-status.php
ingeniumed: Good point, got that change in.

modules/settings/settings.php
ingeniumed: Yip, I've made #60

modules/custom-status/custom-status.php
ingeniumed: Done all this as well

modules/custom-status/custom-status.php
ingeniumed: Actually, this is something that we should remove entirely. If the post type is unsupported, none of our code should load which means we shouldn't even be doing anything.

modules/custom-status/custom-status.php
ingeniumed: Done

modules/shared/php/options-utilities.php
ingeniumed: I had originally kept it because I had tweaked the loaded_once code to use it. But I reverted that code and never ended up tweaking this. Done this as well

modules/custom-status/meta/position-handler.php
ingeniumed: Agreed, I didn't do that yet as it's going to require testing and planning to ensure it doesn't cause any conflicts. I also wanted to use that as a chance to make sure proper GET/GET ALL were in place so those became the default entry points for everything instead of the individual methods.

modules/settings/settings.php
ingeniumed: Yip, done this too. I left the slug as a default in the parameter as the cleanup method that we use in the tests utilizes it.

modules/settings/settings.php
ingeniumed: Thought I fixed that everywhere, but did not looks like.

modules/custom-status/custom-status.php
ingeniumed: I added this back in because without this, the sidebar fails to load. We need to issue that initial status change into draft, after which we can let Gutenberg take over. This is because initially the post type is not available as it's a clean new post.

ingeniumed: This is also why I added pending review as a restricted status because technically that and draft are required by default in core. Us allowing that to be renamed or altered is going to mean extra logic on our end that we need to handle. That is,

- Pending review can be renamed so the label that's used in the custom status registration would need to be tested
- But, the slug itself won't be and will stay the same

Instead, for now I think we should just leave pending review and draft as statuses that cannot be altered or touched in any way. Not saying we don't do it mind you just saying for now we do that and revisit it.

modules/custom-status/custom-status.php
ingeniumed: Given what I said [here](https://github.com/Automattic/vip-workflow-plugin/pull/59#discussion_r1792705172), I'll do that

modules/custom-status/custom-status.php
ingeniumed: https://github.com/Automattic/vip-workflow-plugin/pull/59#discussion_r1792705172

modules/custom-status/custom-status.php
ingeniumed: https://github.com/Automattic/vip-workflow-plugin/pull/59#discussion_r1792705172

modules/custom-status/custom-status.php
```diff
@@ -41,8 +41,8 @@ class CustomStatus {
 	private static $custom_statuses_cache = [];
 
 	public static function init(): void {
-		// Register the taxonomy we use with WordPress core
-		add_action( 'init', [ __CLASS__, 'register_custom_status_taxonomy' ] );
+		// Register the taxonomy we use with WordPress core, and ensure it's registered after editorial metadata
```
ingeniumed, line 44: Did this so that the editorial metadata is included in the custom status properly

modules/custom-status/custom-status.php
```diff
@@ -281,55 +227,60 @@ public function action_admin_enqueue_scripts() {
...
-	 * This function is only necessary until the big refactor is done. After that, it should not be necessary.
+	 * This isn't done anywhere else due to the taxonomies being registered at different times.
+	 * In addition, registering the taxonomies in the wrong order can cause the manage posts page to break
+	 * as well as the default status for a post itself.
 	 *
-	 * @return array $custom_statuses The custom statuses with editorial metadatas
+	 * @return array $custom_statuses The custom statuses with the editorial metadatas included
 	 */
-	private function get_custom_statuses_with_editorial_metadata() {
+	private static function modify_custom_statuses_with_editorial_metadata(): array {
```
ingeniumed, line 258: I made a much more optimized version of the previous method. I also updated the description to mention why it's there. No matter what I did, something broke so I had to do it this way.

If there is a good way to link two taxonomies together to mirror how a foreign key relationship works then we can do that. Otherwise, we need to add this in. 

I think a better option is to stop forcing a UI optimization, and instead handle this how a regular backend talking to a client via a rest api would:

- Add a get all and get by id to CS and EM
- Scrap the server side rendered CS and EM fields
- Have the UI make a call to the get all to render the list, and get by id to render the individual fields as needed
- Lazy load everything to have the UI render quickly and individually make the necessary calls to assemble all the information needed

This would still benefit from the VIP caching logic. It'll be a bit tricky for the block editor, so we can start with the admin and slowly implement it for the block editor.

alecgeatches: I agree that this is a bit weird of a setup. I'm still not sold on REST being the solution, as we're doing the same work but asyncronously and client-side after the page loads. For page state that determines if the save button is disabled, it'd be great to not have to wait for REST calls to determine what to show.

Another idea here could be to pass custom statuses and relevant EM fields separately. Custom statuses have an array of EM IDs, so it wouldn't be hard to have the frontend code in charge of selecting and rendering the relevant fields. It would be cleaner then stuffing data into an existing object, which feels icky here, and still let us avoid the REST roundtrips after load.

alecgeatches: This is still 👍 for now, just thinking about how we'd change this in the future.

tests/modules/custom-status/rest/test-custom-status-endpoint.php
```diff
@@ -17,15 +18,18 @@
...
 
 	/**
-	 * Before each test, ensure default custom statuses are available.
+	 * Before each test, ensure default custom statuses are available and reset all module options.
 	 */
 	protected function setUp(): void {
 		parent::setUp();
 
+		// Reset all module options
+		OptionsUtilities::reset_all_module_options();
```
alecgeatches: If we need to reset options and install for each test, we could put these in `WorkflowTestCase` `setUp()`/`tearDown()` methods.

alecgeatches:
@ingeniumed Per [the test comment above](https://github.com/Automattic/vip-workflow-plugin/pull/59#discussion_r1793793755), I spent time looking into why we needed to sprinkle

```
OptionsUtilities::reset_all_module_options();
CustomStatus::setup_install();
```

around some tests but not others. I found some inconsistencies in our tests, and was able to fix those and simplify our testing code in https://github.com/Automattic/vip-workflow-plugin/pull/59/commits/c69cb671d73a007732ab97c25b6c7edd5b4438eb. The main overview:

- Some of our tests were extending `WP_UnitTestCase`, and some were extending the vanilla PHP `TestCase`. It's hard to find explicit documentation for `WP_UnitTestCase`, but here's what I discovered:
    - The `WP_UnitTestCase_Base` class that `WP_UnitTestCase` extends is [in this repository](https://core.trac.wordpress.org/browser/trunk/tests/phpunit/includes/abstract-testcase.php). From looking at this class, we can see it automatically wraps up everything in a transaction and handles data cleanup. It also clears out WordPress globals and other things that could cause edge cases in tests.
    - These tests were failing because those that specifically extended `WP_UnitTestCase` were getting automatic cleanup and causing our initialized data to disappear.

    - WordPress also has some related documentation for using `WP_UnitTestCase` in [Writing PHP Tests](https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/).
    - I found some setup documentation for integrating WP's classes with composer [in the wp-phpunit/docs repository](https://github.com/wp-phpunit/docs), which is why `bootstrap.php` changed.
- With this information, all tests have been changed to extend `WP_UnitTestCase`. This gives us some advantages:
    - We also get factories for common object types like `user` and `post`. I replaced our custom `create_user()` method with `$this->factory()->user->create()` calls.
 
    - We no longer need to process post/EM field cleanup, because that data is erased on each test run now anyway. Although `WP_UnitTestCase` has an overhead per-test, we get some of that time back by not needing to explicitly run cleanup for individual data.
 
As a side note I noticed that options changes still need to be explicitly cleared, so `WorkflowTestCase` does the reset/setup steps in its `setUp()` so none of our tests need to worry about it.

Overall, this should make our tests more WordPressy, resilient to weird edge-cases, and make it easier for us to add more tests.

tests/modules/custom-status/rest/test-custom-status-endpoint.php
alecgeatches: Addressed in [this comment below](https://github.com/Automattic/vip-workflow-plugin/pull/59#issuecomment-2402886191).

alecgeatches:
I'm going to merge. Thank you for all of the hard work on this PR!

PR #62, opened 2024-10-10
Fix errors on updating the draft and pending status

ingeniumed:
## Description

- Fix an issue introduced with #59 that stopped updates being possible to the draft and pending review statuses, as they were marked as restricted entirely instead of just to their names and slug.
- As part of this, I've also enabled updates to both the draft and pending review status's names. Internally, we reset the slug if it's ever updated so we never need to re-assign posts, etc. We also ensure that the new names are registered correctly.
- I have also updated the validation logic to be a lot more stricter as previously there were some edge cases that could get through.

Deletion of restricted statuses is still not allowed but this would be a good middle ground imo.

## Steps to Test

- Test from scratch that the names of draft and pending review can be changed
- Test also that new statuses with the same name as one existing or a restricted status cannot be made

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -681,7 +681,7 @@ class CustomStatus {
 
 		$old_status_slug = $old_status->slug;
 
-		if ( self::is_restricted_status( $old_status_slug ) && ! self::is_status_banned_from_slug_changes( $old_status_slug ) ) {
+		if ( self::is_restricted_status( $old_status_slug ) || self::is_status_banned_from_slug_changes( $old_status_slug ) ) {
 			// translators: %s: Post status, like "Draft"
 			return new WP_Error( 'restricted', sprintf( __( 'Restricted status (%s) cannot be deleted.', 'vip-workflow' ), $old_status->name ) );
 		}
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -101,6 +101,14 @@ class CustomStatus {
 	 * Also unregisters pending, in case the user doesn't want them.
 	 */
 	public static function register_custom_statuses(): void {
+		global $wp_post_statuses;
+
+		// This will ensure that if the names of the pending and draft statuses are changed, they will
+		// be registered as the new names, rather than the old ones.
+		// They cannot be deleted, so this is only for that.
+		unset( $wp_post_statuses['draft'] );
+		unset( $wp_post_statuses['pending'] );
+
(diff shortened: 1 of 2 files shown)


ingeniumed:
image

Draft and pending statuses can now be renamed with this PR!

modules/custom-status/custom-status.php
```diff
@@ -583,24 +596,23 @@ public static function update_custom_status( int $status_id, array $args = [] ):
...
-		if ( self::is_restricted_status( $old_status->slug )
-		&& (
-			( isset( $args['name'] ) && $args['name'] !== $old_status->name )
-			||
-			( isset( $args['slug'] ) && $args['slug'] !== $old_status->slug )
-		) ) {
-			// translators: %s: Post status, like "Draft"
-			return new WP_Error( 'restricted', sprintf( __( 'Changing the name and slug of a restricted status (%s) is not allowed.', 'vip-workflow' ), $old_status->name ) );
+		// If the name was changed, we need to change the slug unless its banned from slug updates
+		if ( isset( $args['name'] ) && $args['name'] !== $old_status->name && ! self::is_status_banned_from_slug_changes( $old_status->slug ) ) {
```
alecgeatches, line 600: I think we can remove the `! self::is_status_banned_from_slug_changes( $old_status->slug )` check from here, as the code unsets `$args['slug']` if the status is banned below anyway.

modules/custom-status/custom-status.php
```diff
@@ -669,7 +681,7 @@ public static function delete_custom_status( int $status_id ): bool|WP_Error {
 
 		$old_status_slug = $old_status->slug;
 
-		if ( self::is_restricted_status( $old_status_slug ) ) {
+		if ( self::is_restricted_status( $old_status_slug ) && ! self::is_status_banned_from_slug_changes( $old_status_slug ) ) {
```
alecgeatches: This code didn't seem to work for me - I can delete Draft and Pending Review statuses right now, in fact all statuses can be deleted until the last throws an error:

delete-all-statuses

Maybe this is supposed to be:

```suggestion
		if ( self::is_restricted_status( $old_status_slug ) || self::is_status_banned_from_slug_changes( $old_status_slug ) ) {
```

ingeniumed: Yip, accidentally flipped the signs. I tested everything, and then did a bit of cleanup and a bad copy paste went in

modules/custom-status/custom-status.php
ingeniumed: Tbh, I put it there for safety reasons just in case

PR #66, opened 2026-05-04
Fix PHP 8 fatal error when editing unsupported post type in block editor

saqibsarwar:
## Problem

When creating or editing a post of a post type that is **not registered with VIP Workflow**, the block editor throws a PHP 8 fatal error:

```
PHP Fatal error: Uncaught Error: Attempt to modify property "meta" on array
in modules/custom-status/custom-status.php:278
```

**Stack trace:**
```
CustomStatus::modify_custom_statuses_with_editorial_metadata()
CustomStatus::load_scripts_for_block_editor()
WP_Hook->apply_filters()
do_action('enqueue_block_editor_assets')
```

## Root Cause

`get_custom_statuses()` correctly detects that the post type is unsupported and falls back to `get_core_statuses()`. However, `get_core_statuses()` returns plain associative arrays, while `modify_custom_statuses_with_editorial_metadata()` assumes every item is a `WP_Term` object and writes to `$status->meta[...]`. PHP 8 treats this as a fatal error.

## Fix

Add an `is_object()` guard in the loop inside `modify_custom_statuses_with_editorial_metadata()` to skip non-object items. Core status arrays carry no editorial metadata so skipping them is semantically correct — no behaviour change for supported post types.

## Testing

1. Register a post type that is **not** added to VIP Workflow's supported post types.
2. Navigate to **New Post** for that post type in the block editor.
3. Before this fix: PHP fatal error, white screen.
4. After this fix: block editor loads normally.

### DIFF
--- a/modules/custom-status/custom-status.php
+++ b/modules/custom-status/custom-status.php
@@ -268,8 +268,15 @@ class CustomStatus {
 
 		$custom_statuses = self::get_custom_statuses();
 
-		// Add the required editorial metadata to the custom statuses for UI purposes
+		// Add the required editorial metadata to the custom statuses for UI purposes.
+		// When the current post type is unsupported, get_custom_statuses() returns the result of
+		// get_core_statuses(), which are plain arrays rather than WP_Term objects. Skip those since
+		// they carry no editorial metadata, and attempting object property access on an array is a
+		// fatal error in PHP 8.
 		foreach ( $custom_statuses as $status ) {
+			if ( ! is_object( $status ) ) {
+				continue;
+			}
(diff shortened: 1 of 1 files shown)


maxschmeling (approved):

alecgeatches:
@saqibsarwar Thank you for the fix! It seems like maybe `get_custom_statuses()` should be modified to return a consistent type, e.g. everything is put into a `WP_Term` object OR associative array so it doesn't matter if `get_core_statuses()` is used. That said, this plugin is not under active development and if this unblocks you for something I'm happy to get it in fast.

If you do run into this issue again with another `get_custom_statuses()` consumer then we should probably implement the idea above for consistency. Going to approve for now!

alecgeatches (approved):

